PR #16, opened 2016-09-03
Implement proper test finder for GoogleTest

AlexDenisov:


lib/Driver.cpp
```diff
@@ -84,7 +84,6 @@ std::vector<std::unique_ptr<TestResult>> Driver::Run() {
         ExecutionResult R = Runner.runTest(BorrowedTest, ObjectFiles);
         assert(R.Status != ExecutionStatus::Invalid && "Expect to see valid TestResult");
 
-        /// FIXME: Check if it's legal or not
```
AlexDenisov: It is legal

PR #17, opened 2016-09-04
1 forked process per mutation

stanislaw:


lib/Driver.cpp
```diff
@@ -63,25 +63,34 @@ std::vector<std::unique_ptr<TestResult>> Driver::Run() {
...
+
+          auto Mutant = Compiler.CompilerModule(Testee->getParent());
+          ObjectFiles.push_back(Mutant.getBinary());
 
-        /// TODO: here the clone of Testee->getParent() will be used very soon instead.
-        /// For now we are applying mutation to the module same as of mutation point.
-        MutationPoint->applyMutation(Testee->getParent());
+          /// No longer need to revert mutations since forked child dies:
+          /// ~~Rollback mutation once we have compiled the module~~
+          /// MutationPoint->revertMutation();
```
mbj: Also rollback of mutations may not be possible, since executing the mutation on a language with global side effects can leak state from one mutation into the next one. So +1 for this change. (but I'd remove commented code).

lib/ForkProcessSandbox.cpp
```diff
@@ -0,0 +1,48 @@
...
+  else if (ChildPid == 0) {
+    Func(SharedResult);
+    exit(0);
+  }
+
+  else {
+    /// TODO: Handle at least 3 cases:
+    /// 1) Child process crashes
+    /// 2) Child process exits 1
+    /// 3) Child process hangs
```
mbj: In mutant I decided to explicitly NOT handle this case from the mutation engine, and decided the test frameworks itself must terminate. The "child process hangs" case is de-facto a halting problem, its inherently unsolvable and only can be mitigated. Timeouts IMO should for that reason than be more close to the source of the intent halting: The test framework.

Also this avoids having to maintain 3 states in reporting, as a "hanged" child process cannot be easily counted as kill or alive.

lib/Driver.cpp
```diff
@@ -63,25 +63,34 @@ std::vector<std::unique_ptr<TestResult>> Driver::Run() {
...
+      *SharedResult = Runner.runTest(Test.get(), ObjectFiles);
+    });
+
     auto BorrowedTest = Test.get();
     auto Result = make_unique<TestResult>(ExecResult, std::move(Test));
 
     for (auto Testee : Finder.findTestees(BorrowedTest, Ctx)) {
       auto ObjectFiles = AllButOne(Testee->getParent());
       for (auto &MutationPoint : Finder.findMutationPoints(MutationOperators, *Testee)) {
+        ExecutionResult R = Sandbox.run([&](ExecutionResult *SharedResult){
```
mbj: I do not know if the test framework is "pre booted" in mutang, I just want to mention at this point that fork based sandboxing (I call it isolation in mutant) comes at a non obvious cost: If killing the mutation has a one time setup cost, like scanning test files, or anything that would normally happen *once* for N kills, now gets payed *per* kill. As this one time setup work has to be re-done per sandbox.

For this reason in mutant I try to trigger all "one time" work before attempting mutation kills, to inherit that one time work with the fork.

If this strategy is even required in mutangs case I should note here that there may need to be a hook so test frameworks can re-connect the database to unshare FDs (assuming the test framework on performing the one-time work would create such).

mbj:
@stanislaw I was watching this repository for a while, as I like mutation testing a lot in general. And I implemented a "sucessful" (as in more than expected amounts of people use it) engine for Ruby. I do like to talk about the various implementation trade-offs a lot, so I just commented some of my thoughts.

I hope its useful, and or may at least be the nucleus for some interesting discussion where I can learn something about other ecosystems.

//cc @AlexDenisov

lib/Driver.cpp
AlexDenisov: At this point  we work only with LLVM IR split into so-called modules.
To run a test the IR need to be compiled into an object file, once we get an object file we can change the module, compile it again and get another object file, i.e. mutated program.
So no state is rolled back.
The idea behind is related to premature optimization: I did the rollback just to not make a copy of a module every time we want to make a mutation. Though the mechanism is not very trivial, so we decided to skip it for now and to move the responsibility for copies to OS: when you do a fork the child inherits all the parent's memory in read-only mode, then when you try to modify parent's memory it implicitly gets copied into child's address space. 
I.e. we get the copying 'for free'. I hope this hypothesis works :)

lib/ForkProcessSandbox.cpp
AlexDenisov: We cannot determine if program hangs or not, however this is not the goal.
We just want to stop hanging mutant after some timeout.

C* languages have a nice structure, called for-loop:
```
(for int i = 0; i < 5; i++) {
 /// do stuff
}
```
Here, if we replace `++` with `--` - the program hangs. There is a workaround: to not apply such a mutation in case if it's an index from a for-loop.
However, I imagine many more exceptions like this. It's better to see whether program hangs because of 'wrong' mutation, than to ignore this problem.

We can make many assumptions regarding potential hanging mutants an workaround them in advance, but imo it's better to face them in reality and then solve.

How did you overcome this problem?

lib/Driver.cpp
AlexDenisov: All the setup work already done at this point. We just deferring mutation and mutant execution to a child process to gain implicit module copy (mentioned above) and to prevent ourselves from hanging/crashing.

AlexDenisov: Hm, I may get you wrong.

For each test/mutant run we 'boot' test framework from scratch (running all static constructors this case), and after the run we shut it down (running all destructors).
I'm afraid we can't do it other way because of a state. 
GoogleTest particularly does lots of things using static variables.

AlexDenisov:
@mbj it is extremely helpful! 👍 
Please, feel free to ask/comment/say anything you have in mind.

lib/Driver.cpp
mbj: > The idea behind is related to premature optimization: I did the rollback just to not make a copy of a module every time we want to make a mutation. Though the mechanism is not very trivial, so we decided to skip it for now and to move the responsibility for copies to OS: when you do a fork the child inherits all the parent's memory in read-only mode, then when you try to modify parent's memory it implicitly gets copied into child's address space. 
> I.e. we get the copying 'for free'. I hope this hypothesis works :)

Even when you can perfectly rollback the semantic change, if you do not have isolation and you executed a mutation behavior of executing the mutation could leak into the next mutations. Example: Leaking FDs on tests that use them, caused by mutating away the close call. And others. 

Hence attempting to re-use the same VM on mutating not guaranteed to be pure code there can be any sorts of nasty "escaped effects" (not escaped code inserted) infecting next tests.

Sandboxing by default is much more safe. A mutation testing system for Haskell I'm working on can exploit the property of guaranteed pureness to increase efficiency as the sandbox can deterministically be skipped for some mutations.

lib/ForkProcessSandbox.cpp
mbj: > We can make many assumptions regarding potential hanging mutants an workaround them in advance, but imo it's better to face them in reality and then solve.
> 
> How did you overcome this problem?

Forcing the test frameworks that would execute the tests against such infinite runtime mutations (thats the name I came up with) to have a timeout. But not the mutation testing engine itself.

The idea is that the the following done by hand, should be quivalent to what the mutation testing engine does:

* Insert the code by hand
* Run the subset of the tests the mutation testing engine would run

can result in the same "kill reason" for being observable. If the test framework has a timeout, the user can reproduce the "kill reason" by hand, but if the mutation testing engine on "top of the test framework" has a timeout, the user cannot realistically reproduce the same kill reason, on more complex cases this can lead to significant amounts of time wasted.

Hence my call for: Tests should enforce timeouts, not mutation testing engines.

lib/Driver.cpp
mbj: > For each test/mutant run we 'boot' test framework from scratch (running all static constructors this case), and after the run we shut it down (running all destructors).
> I'm afraid we can't do it other way because of a state. 
> GoogleTest particularly does lots of things using static variables.

Yes, exactly. In mutants case (Ruby) performing the equivalent of static initializers (the gazillions of DSL call sides) can if shared as forced to be done "once" before the N forks significantly increase performance.

Hence I was just mentioning that once you observe "redunant work" being "re-done" per fork, you may be able to force this work to be inherited. Removing the time spend inside the fork significantly.

mbj:
@AlexDenisov I replied in-line.

I like this discussion, keep it coming.

lib/Driver.cpp
AlexDenisov: > the mutation could leak into the next mutations

This is a very good point. Such a perfect hotbed for nasty bugs.

lib/ForkProcessSandbox.cpp
AlexDenisov: > Hence my call for: Tests should enforce timeouts, not mutation testing engines.

I wish we could rely on it, I'm afraid we will have to implement the timeouts just for our own safety.
But we will see, I'm not 100% sure. 
@stanislaw ?

lib/Driver.cpp
AlexDenisov: Hm, I didn't think about this opportunity. We will definitely need to investigate.
@stanislaw let's keep this in mind and check how the system behaves in such case.

lib/ForkProcessSandbox.cpp
mbj: > I wish we could rely on it, I'm afraid we will have to implement the timeouts just for our own safety.
> But we will see, I'm not 100% sure.

Maybe you want both, but consider a timeout hit from the mutation testing engine a bug. Assumption: If the test framework could not enforce a timeout, when executed from mutation testing, something is really wrong with the integration.

This is at least the path mutant will do in future, test timeouts in ruby are possible but not 100% solid.

PR #58, opened 2016-12-14
Add RemoveVoidFunction operator.

stanislaw:


lib/MutationOperators/RemoveVoidFunctionMutationOperator.cpp
```diff
@@ -0,0 +1,101 @@
...
+    /// How it can be that there is no a called function?
+    if (Function *calledFunction = callInst->getCalledFunction()) {
+      if (calledFunction->getName().startswith("llvm.dbg.declare")) {
+        return false;
+      }
+
+      if (calledFunction->getName().endswith("D1Ev") || calledFunction->getName().endswith("D2Ev") ||
+          calledFunction->getName().contains("C1E") || calledFunction->getName().contains("C2E")) {
+        return false;
+      }
```
AlexDenisov: 1. It should not be here since it's very C++ specific
2. It's not implemented properly. It will probably also filter some valid methods, not just ctors/dtors.

Could you please at least add some todo/fixme here?

PR #63, opened 2016-12-16
Use LLVM Yaml Traits for config

JDevlieghere:
- Use [LLVM's YAML I/O](http://llvm.org/docs/YamlIO.html) Library for parsing the configuration file.
 - Consistent options: everything now uses an underscore (e.g. dry_run instead of DryRun).
 - Config is no longer a unique pointer. If something fails you end up with the default config. If this is not the desired behavior, we could pass the config by reference and have the function return a bool or use as a 3rd option we could use the llvm::Expect<T> type.

### DIFF
--- a/include/Config.h
+++ b/include/Config.h
@@ -1,10 +1,23 @@
 #pragma once
 
+#include "llvm/Support/YAMLTraits.h"
+
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


stanislaw:
Looks very good to me. @AlexDenisov ?

Issue #14, opened 2016-08-24
Logging: add logger to see where we are while running the program.


AlexDenisov:
At the moment if I run program against big-ish test suite (~700 modules) I don't get any feedback, i.e. what the program does at the moment.

PR #65, opened 2016-12-18
Preserve testee call paths to make them available later for reporting.

stanislaw:


lib/Driver.cpp
```diff
@@ -46,8 +48,10 @@ using namespace std::chrono;
 /// Each result contains result of execution of an original test and
 /// all the results of each mutant within corresponding MutationPoint
 
-std::vector<std::unique_ptr<TestResult>> Driver::Run() {
+std::unique_ptr<Result> Driver::Run() {
   std::vector<std::unique_ptr<TestResult>> Results;
+  std::vector<std::unique_ptr<Testee>> allTestees;
+  allTestees.reserve(100000);
```
AlexDenisov: Why is that?

lib/Result.cpp
```diff
@@ -0,0 +1,67 @@
...
+  auto mutationPoint = mutationResult->getMutationPoint();
+  Instruction *instruction = dyn_cast<Instruction>(mutationPoint->getOriginalValue());
+  std::string fileName = instruction->getDebugLoc()->getFilename();
+  std::string line = std::to_string(instruction->getDebugLoc()->getLine());
+
+  ///Function *function = instruction->getFunction();
+  ///assert(function && "Expected function");
+  ///std::string functionName = function->getName().str();
+
+  snprintf(formatBuffer, sizeof(formatBuffer), "-- %s:%s\n", fileName.c_str(), line.c_str());
```
AlexDenisov: Why `snprintf`?

AlexDenisov: I'd make this function to return the path as an array of strings, and then format accordingly inside of reporter. This way we may save it in different formats (e.g. in pre-formatted and in csv form), and experiment which one works better.

lib/Driver.cpp
stanislaw: can be removed, copy and paste artefact for vector concatenation I didn't know how to do in CPP 😝

lib/Result.cpp
stanislaw: > Why snprintf?

I didn't know how to std::string with format better than that way.

> I'd make this function to return the path as an array of strings, and then format accordingly inside of reporter. This way we may save it in different formats (e.g. in pre-formatted and in csv form), and experiment which one works better.

How should I then write array of strings to SQLite?

stanislaw: > I didn't know how to std::string with format better than that way.

@JDevlieghere maybe you know this?

JDevlieghere: The idiomatic C++ approach is to use streams for this. I was already planning on tackling #14 and implementing a small logger class on top of LLVM's streams, which we could use here.

AlexDenisov: > How should I then write array of strings to SQLite?

As you do it now, but computing the pre-formatted string later, giving a chance to use another format (for example csv).

lib/Driver.cpp
```diff
@@ -203,10 +218,10 @@ void Driver::debug_PrintMutationPoints() {
 
   for (auto &Test : Finder.findTests(Ctx)) {
     outs() << Test->getTestName() << "\n";
-    for (auto testee : Finder.findTestees(Test.get(), Ctx, Cfg.getMaxDistance())) {
-      auto MPoints = Finder.findMutationPoints(Ctx, *(testee.first));
+    for (auto &&testee : Finder.findTestees(Test.get(), Ctx, Cfg.getMaxDistance())) {
```
JDevlieghere: I don't think you really need the universal reference here. I think it's safer to take the lvalue reference and "move" it if you want to transfer ownership. Now it's hard to tell if ownership it transferred in the first place and where that happens.

stanislaw:
@AlexDenisov @JDevlieghere I have made the changes, please review.

stanislaw:
I have made them 1 commit per change so hopefully clear enough those changes are.

lib/SQLiteReporter.cpp
```diff
@@ -215,6 +206,24 @@ void Mutang::SQLiteReporter::reportResults(const std::unique_ptr<Result> &result
   outs() << "Results can be found at '" << databasePath << "'\n";
 }
 
+#pragma mark -
+
+std::string SQLiteReporter::getCallerPathAsString(std::vector<std::string> callerPath) {
```
JDevlieghere: Good idea to extract this in a separate function! Note though that you're passing the vector by value here, which means that you're making a copy. Better pass it by const reference.

stanislaw: https://github.com/mull-project/mull/pull/65/commits/2e44675c2d0d43fc10527ddf6d8a4206e970a8c1

Correct?

JDevlieghere: You can make the reference const as you're not changing the vector, I think.

stanislaw: This is what goes beyond my limited knowledge of C++. Should const go like:

```
std::string getCallerPathAsString(std::vector<std::string> const &callerPath);
```

or 

```
std::string getCallerPathAsString(const std::vector<std::string> &callerPath);
```

?

JDevlieghere: That's alright, that's one of the benefits of code reviews. :-) To answer your question, both are equivalent semantically. In my experience the second type is most commonly used, so I'd recommend that one.

stanislaw: Xcode denies `push_back()` on both ways. I have made a change with the second form.

lib/Result.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  std::vector<std::string> callerPath;
+
+  /// Currently we don't print functions, only "file:line". Later we will and
+  /// for that we will have to demangle them.
+
+  /// Last path component: mutation point itself.
+  auto mutationPoint = mutationResult->getMutationPoint();
+  Instruction *instruction = dyn_cast<Instruction>(mutationPoint->getOriginalValue());
+  std::string fileName = instruction->getDebugLoc()->getFilename();
```
JDevlieghere: this can be `const std::string`

lib/Result.cpp
```diff
@@ -0,0 +1,61 @@
...
+  std::string line = std::to_string(instruction->getDebugLoc()->getLine());
```
JDevlieghere: const

lib/Result.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  ///Function *function = instruction->getFunction();
+  ///assert(function && "Expected function");
+  ///std::string functionName = function->getName().str();
+
+  std::stringstream mpComponentAsStream;
+  mpComponentAsStream << fileName.c_str();
```
JDevlieghere: no need for the `c_str()`

lib/Result.cpp
```diff
@@ -0,0 +1,61 @@
...
+  mpComponentAsStream << ":";
+  mpComponentAsStream << line.c_str();
+  std::string mpComponent = mpComponentAsStream.str();
```
JDevlieghere: You're not reusing this string so you can save an allocation by doing `callerPath.push_back(mpComponentAsStream.str())` directly.

lib/Result.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  callerPath.push_back(mpComponent);
+
+  // The following loop stops on test as it does not have a caller.
+  do {
+    Instruction *instruction = currentTestee->getCallerInstruction();
+
+    std::string fileName = instruction->getDebugLoc()->getFilename();
```
JDevlieghere: same as above, can be const.

lib/Result.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::string line = std::to_string(instruction->getDebugLoc()->getLine());
```
JDevlieghere: also const.

lib/Result.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    std::stringstream callerComponentAsStream;
+    callerComponentAsStream << fileName.c_str();
```
JDevlieghere: same as above, no `c_str()` necessary.

lib/Result.cpp
```diff
@@ -0,0 +1,61 @@
...
+    callerComponentAsStream << ":";
+    callerComponentAsStream << line.c_str();
+    std::string callerComponent = callerComponentAsStream.str();
+
+    callerPath.push_back(callerComponent);
```
JDevlieghere: same as above, no need for the temporary string.

lib/SimpleTest/SimpleTestFinder.cpp
```diff
@@ -44,42 +61,137 @@ std::vector<std::unique_ptr<Test>> SimpleTestFinder::findTests(Context &Ctx) {
...
+
+          /// The code below is not actually correct
+          /// For each C++ constructor compiler can generate up to three
+          /// functions*. Which means that the distance might be incorrect
+          /// We need to find a clever way to fix this problem
+          ///
+          /// * Here is a good overview of what's going on:
+          /// http://stackoverflow.com/a/6921467/829116
+          ///
+          traversees.push(new Testee(definedFunction,
```
JDevlieghere: This smells like a memory leak, you're allocating testees on the stack, yet they are never deleted. Why not make it a vector of `unqiue_ptr`'s?

lib/SimpleTest/SimpleTestFinder.cpp
```diff
@@ -44,42 +61,137 @@ std::vector<std::unique_ptr<Test>> SimpleTestFinder::findTests(Context &Ctx) {
...
+                                     callInstruction,
+                                     traversee,
+                                     mutationDistance + 1));
         }
       }
+    }
 
+    traversees.pop();
+    if (traversees.size() == 0) {
```
JDevlieghere: `::empty()`

lib/SimpleTest/SimpleTestFinder.cpp
stanislaw: I will address all of the comments tomorrow. For now one question about this: later this allocated testee gets pushed to `std::vector<std::unique_ptr<Testee>> testees;` via

```
std::unique_ptr<Testee> uniqueTestee(traversee);
testees.push_back(std::move(uniqueTestee));
```

doesn't that mean that this allocation will be managed by vector of unique pointers?

That's what my assumption is about how unique pointers work. That's not the only place in the code where I and Alexey exploit this behaviour. Are we wrong?

AlexDenisov: I think @JDevlieghere missed this point, for me the code looks correct-ish.
Although, there is a memory leak indeed, since we make unique pointers out of traversees conditionally:
```
if (traverseeFunction->getParent() != testBodyModule) {
  std::unique_ptr<Testee> uniqueTestee(traversee);
  testees.push_back(std::move(uniqueTestee));
}
```

stanislaw: Good points! I'll make the changes today :)

stanislaw: Hm, that's a bit hard because the of the first so-called `topLevelTestee` which I am already pushing to `testees` before we start working with traversees in the loop. Any ideas?

stanislaw: So we have two parallel collections of unique pointers if we follow the way you @JDevlieghere suggested. I just don't understand what's the right way to do that.

JDevlieghere: @stanislaw Is it alright if I look at it tomorrow?

stanislaw: @JDevlieghere I have pushed all the changes regarding your comments. This one with testees/traversees is the last one. Do you have a good idea on how to unique_pointer all of this? :)

JDevlieghere: @stanislaw I will clone this branch and give it a try. It's always easier to do this with some context. :-)

stanislaw: I know how to do it in a hacky way, but maybe you know something better.

JDevlieghere: Can you join the mull slack? Might be easier to discuss some stuff.

PR #69, opened 2016-12-21
Add Logger

JDevlieghere:
- Added logger class
 - Replaced some printf's with new logger
 - Replaced uses of outs() and errs() with new logger


JDevlieghere:
Thanks @stanislaw! I'll wait for feedback from @AlexDenisov before I merge it in.

Issue #66, opened 2016-12-18
Find a way to merge algorithms of Simple and Google Test finders into one.


stanislaw:
Currently both algorithms are almost identical. Also one-sided development of Google Test may result in too opinionated class, which could (and should) be generic.

AlexDenisov:
Fixed via #218.

PR #71, opened 2016-12-22
Experiment: SimpleTestFinder::findTestees all unique_ptr

stanislaw:


stanislaw:
@JDevlieghere this is what I wanted to show you however I couldn't remove the "convoluted" parts properly which you did.

stanislaw:
I really don't like that the same algorithm exists in both finders: `SimpleTestFinder` and `GoogleTestFinder`. We need to do #66.

Also currently there are no unit tests for `GoogleTestFinder` :(

PR #74, opened 2016-12-24
Remove mutation engine

stanislaw:
This should be merged after #73 .


unittests/MutationPointTests.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+    if (Instr.getName().equals(Name)) {
+      return &*I;
+    }
+  }
+
+  return nullptr;
+}
+
+TEST(MutationEngine, SimpleTest_AddOperator_applyMutation) {
```
AlexDenisov: Could you please also change the name here?

stanislaw: fair pedantic point 😄

PR #78, opened 2016-12-29
Tweak debug log

AlexDenisov:


lib/Driver.cpp
```diff
@@ -115,7 +115,7 @@ std::unique_ptr<Result> Driver::Run() {
 
       auto MPoints = Finder.findMutationPoints(Ctx, *(testee->getTesteeFunction()));
       if (MPoints.empty()) {
-        Logger::debug() << "\t\no mutation points, skipping.\n";
+        Logger::debug() << "\t\tno mutation points, skipping.\n";
```
JDevlieghere: Instead of using tabs, you can tell the different loggers to indent with a given numer of spaces using for example `Logger::debug().indent(8) << "message"`

AlexDenisov: Nice, I didn't know. I will update it all over the place later. I'm also adding a bit more useful information.

Issue #53, opened 2016-11-22
SQLiteReporter: escape symbols properly.


stanislaw:
```
Results can be found at '/Users/stanislaw/Projects/LLVM/BuildXcode/Debug/bin/1479765000.sqlite'
Cannot execute 'INSERT INTO execution_result VALUES ('0','0','','LLVM ERROR: Program used external function '__ZTIN4llvm20SectionMemoryManagerE' which could not be resolved!
');'
```

Quolyk:
How can I reproduce this? SQLiteReporter test seems normal.

stanislaw:
@Quolyk can be that it has gone. At least I have completely lost original context. Will close :)

AlexDenisov:
@Quolyk @stanislaw it's not gone.
There are two places where we insert `execution_result` into a database: around lines 79 and 155.
We do it properly only In the second case.

stanislaw:
@Quolyk are you willing to make a contribution? We are happy to help if you are :)

You could extend https://github.com/mull-project/mull/blob/master/unittests/SQLiteReporterTest.cpp with a test that would write the failing line from:

```
Cannot execute 'INSERT INTO execution_result VALUES ('0','0','','LLVM ERROR: Program used external function '__ZTIN4llvm20SectionMemoryManagerE' which could not be resolved!
```

P.S. This is how I would do it :)

Quolyk:
@stanislaw I want to make a contribution :) Thanks for the help.

Quolyk:
How can I create such test?
I suppose I should make `ExecutionResult` that produces 
`'LLVM ERROR: Program used external function '__ZTIN4llvm20SectionMemoryManagerE' which could not be resolved!`?
Btw, there are no `MutationResults` in `TestResult`

AlexDenisov:
You can try to construct an execution result with `stderrOutput` equal to `'LLVM ERROR: Program used external function '__ZTIN4llvm20SectionMemoryManagerE' which could not be resolved!` and attempt to save it.

stanislaw:
Thanks @Quolyk !

PR #80, opened 2017-01-12
SQLiteReporter tests

Quolyk:
#53

### DIFF
--- a/lib/SQLiteReporter.cpp
+++ b/lib/SQLiteReporter.cpp
@@ -46,6 +46,16 @@ void sqlite_exec(sqlite3 *database, const char *sql) {
   }
 }
 
+void sqlite_step(sqlite3 *database, sqlite3_stmt *stmt) {
+  int result = sqlite3_step(stmt);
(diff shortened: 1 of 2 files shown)


Quolyk:
I deleted quotes from `'__ZTIN4llvm20SectionMemoryManagerE'` to correctly write to db. I don't now if I'm on the right way

AlexDenisov:
The right way would be to use the same approach as used in another place: not putting string directly into sql query, but using `?` + prepare statement + bind_text.
Please look at `insertMutationExecutionResultSQL` for a reference.

Quolyk:
I guess, now it's working

AlexDenisov:
Thanks a lot, @Quolyk. Could you please squash commits before the merge?
https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

Quolyk:
@AlexDenisov, Done

AlexDenisov:
You've solved really annoying problem, thanks a lot for your effort :)

PR #88, opened 2017-01-22
Update code to work against newer LLVM.

stanislaw:
- Updates Mull to work against this LLVM commit:

```
commit dc30ded6fb9526aba31a86b97e8f69fc9baece00
Author: Justin Lebar <jlebar@google.com>
Date:   Wed Jan 18 00:29:53 2017 +0000
```

- Small tweak of GoogleTestFinder was needed:

class.testing::TestInfo functions are no longer under sequential type but pointer type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ cd ~/Projects/LLVM
 
 git clone http://llvm.org/git/llvm.git
 cd llvm
-git checkout 106bb1fe2db88fe77458ec594c461c8428588dab
+git checkout dc30ded6fb9526aba31a86b97e8f69fc9baece00
 
 git clone https://github.com/mull-project/mull.git projects/mull
 
@@ -37,5 +37,5 @@ Update to LLVM's trunk should be a recurrent task, but first we need to get
 experience of living in a downstream world.
 
 Current code is being developed against LLVM commit:
-[106bb1fe2db88fe77458ec594c461c8428588dab](https://github.com/llvm-mirror/llvm/commit/106bb1fe2db88fe77458ec594c461c8428588dab).
+[dc30ded6fb9526aba31a86b97e8f69fc9baece00](https://github.com/llvm-mirror/llvm/commit/dc30ded6fb9526aba31a86b97e8f69fc9baece00).
 
--- a/lib/GoogleTest/GoogleTestFinder.cpp
+++ b/lib/GoogleTest/GoogleTestFinder.cpp
@@ -98,12 +98,12 @@ std::vector<std::unique_ptr<Test>> GoogleTestFinder::findTests(Context &Ctx) {
         continue;
       }
 
-      Type *SeqTy = Ty->getSequentialElementType();
-      if (!SeqTy) {
+      Type *globalType = Ty->getPointerElementType();
+      if (!globalType) {
         continue;
       }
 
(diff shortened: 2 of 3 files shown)


JDevlieghere:
Maybe we should aim to link against a release branch of LLVM rather than a "random" commit?

stanislaw:
@JDevlieghere, sorry I don't know what that is. What does it mean to link against a release branch?

AlexDenisov:
@JDevlieghere I think that we may need some trunk's features in the future (i.e. JIT support for ObjC). However, we can have some process in place, e.g. update every first day of a month.

JDevlieghere:
@stanislaw The LLVM contains several tagged branches that correspond to releases. For example `release_40` is the fork that will ultimately be released as LLVM 4.0. Once it's forked no more API changes are allowed, only bugfixes, so you can build against a stable version. The problem with a random commit is that stuff might be broken, which you wouldn't know until buildbot fails. I'm speaking from experience. :-) 

Anyway, if @AlexDenisov wants to stay as close to trunk as possible, that's fine for me too!

stanislaw:
Thanks for the explanation @JDevlieghere . I would actually stick to stable versions but let's let @AlexDenisov to decide.

JDevlieghere:
You're welcome! I don't have a strong opinion though, it was just a suggestion.

AlexDenisov:
Oh, I misunderstood you @JDevlieghere. I thought you suggest sticking to some stable (6-month old) release.
Do you know how often release branches updated?

stanislaw:
Now I am more convinced that we should rely on stable branches.

@Quolyk is working with GoogleTest's IR fixtures and I am on Rust's IR. We are both hitting the same kind of error when we are parsing IR:

```
test: <string>:7237:187: error: invalid field 'variable'
!1526 = distinct !DIGlobalVariable(name: "test_info_", linkageName: "_ZN14Hello_sup_Test10test_info_E", scope: !0, file: !1527, line: 4, type: !1528, isLocal: false, isDefinition: true, variable: %"class.testing::TestInfo"** @_ZN14Hello_sup_Test10test_info_E, declaration: !2817)
```

Recently IR stuff was changed so that `4.0` is backward incompatible. One example:

```
commit 7b500b4bdf40cb40cb33bdcf5faf900db4930824
Author: Adrian Prantl <aprantl@apple.com>
Date:   Tue Dec 20 02:09:43 2016 +0000

    [IR] Remove the DIExpression field from DIGlobalVariable.
    
    This patch implements PR31013 by introducing a
    DIGlobalVariableExpression that holds a pair of DIGlobalVariable and
    DIExpression.

    ...
```

brew is still `3.9.1` and nightly Rust is `3.9` so at this point we are blocked unless we downgrade. Not sure though if we should wait until brew, Rust and whatever else updates to `4.0` or downgrade to `3.9`. Thoughts, @AlexDenisov ?

TIL: 👍 for working against stable branches!

AlexDenisov:
I'm fine with sticking to release branches. In case we would need something from the trunk, we can find a workaround.

stanislaw:
I will try to downgrade to 3.9 and see if LLVM IR works good. If it does that for me and @Quolyk I'll likely send a downgrade PR with 4.0 stuff commented out.

PR #89, opened 2017-01-26
Make Mutation Operators configurable (with YAML issue fixed)

stanislaw:


stanislaw:
cc @JDevlieghere are you familiar with this weird behaviour of YAML traits library?

```
+    mutationOperators(
+      // Yaml::Traits stops reading mutation_operators from config.yaml (BUG?)
+      // if these 3 default operators are set here. So leaving out the empty {}
+      // {
+      //   AddMutationOperator::ID,
+      //   NegateConditionMutationOperator::ID,
+      //   RemoveVoidFunctionMutationOperator::ID
+      // }
+    ),
```

stanislaw:
cc @Quolyk

stanislaw:
Lesson learned: don't write test that executes custom values that are the same as default values as they might get shadowed. I was reading 3 operators in a unit tests and they were not working but being shadowed by default values. 😥

lib/Driver.cpp
```diff
@@ -204,12 +213,29 @@ std::vector<std::unique_ptr<MutationOperator>> Driver::mutationOperators
...
+Driver::defaultMutationOperators() {
+  std::vector<std::unique_ptr<MutationOperator>> operators;
+
+  operators.emplace_back(make_unique<AddMutationOperator>());
+  operators.emplace_back(make_unique<NegateConditionMutationOperator>());
+  operators.emplace_back(make_unique<RemoveVoidFunctionMutationOperator>());
+
+  return operators;
+}
+
```
AlexDenisov: This is good. We definitely need to have a default option.

PR #114, opened 2017-02-15
CMake: make Mull an out-of-LLVM-tree project.

stanislaw:


stanislaw:
The targets: `Mull`, `MullUnitTests`, `mull-driver` work. I need to do some cleanups but overall it seems to be working perfectly.

stanislaw:
@JDevlieghere 

I am going to move all platform-specific settings to a custom CMake-toolchain file so that one can configure his own toolchain to build on Linux.

I would appreciate a feedback from you about this, expecially maybe you are aware of more idiomatic CMake tricks to make this happen.

stanislaw:
@AlexDenisov @JDevlieghere I am merging this. Would be great if you could test the `Getting started` instructions for OSX.

PR #130, opened 2017-02-26
Cleanup: remove deprecated method applyMutation.

stanislaw:
- Makes more explicit name: `cloneModuleAndApplyMutation` and moves compilation to be outside of MutationPoint class.
- Fixes libc test fixture identifier.
- A couple of naming improvements in unit tests.

### DIFF
--- a/lib/MutationOperators/AddMutationOperator.cpp
+++ b/lib/MutationOperators/AddMutationOperator.cpp
@@ -202,7 +202,7 @@ llvm::Value *AddMutationOperator::applyMutation(Module *M,
   Instruction *replacement = BinaryOperator::Create(Instruction::Sub,
                                                     binaryOperator->getOperand(0),
                                                     binaryOperator->getOperand(1),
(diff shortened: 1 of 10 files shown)


lib/MutationOperators/AddMutationOperator.cpp
```diff
@@ -202,7 +202,7 @@ llvm::Value *AddMutationOperator::applyMutation(Module *M,
   Instruction *replacement = BinaryOperator::Create(Instruction::Sub,
                                                     binaryOperator->getOperand(0),
                                                     binaryOperator->getOperand(1),
-                                                    binaryOperator->getName());
+                                                    "sub");
```
AlexDenisov: It was on purpose. What's your point? :)

stanislaw: Point was to make mutations explicit. I also changed it to make my unit test more explicit but no specific point I'm making. Reverted.

PR #133, opened 2017-03-01
Test filter works in RustTestFinder.

stanislaw:


lib/Rust/RustTestFinder.cpp
```diff
@@ -168,6 +168,13 @@ std::vector<std::unique_ptr<Test>> RustTestFinder::findTests(Context &Ctx) {
 
       auto testFunction = dyn_cast<Function>(testPointerStruct->getOperand(1));
 
+      if (testsToFilter.size() > 0 &&
```
AlexDenisov: `!empty()`

PR #145, opened 2017-04-07
Auto-test suite: Mull tests itself.

stanislaw:


mbj:
I do not have the time to dive into the details, but for curiosity: Does this add the ability to mutation test the mutation testing engine? 

And you did make sure that you do mutate the implementation in a way the mutated mutation testing engine instance is NOT used to analyse the mutation kill, so there is strong isolation between: Mutation testing engine as subject of coverage testing, and mutation testing engine as executor? If so: How did you archive this?

Before mull I only knew mutant to have this self covering feature, hence I'm curious on how the other engines are catching up.

lib/SQLiteReporter.cpp
```diff
@@ -293,7 +293,9 @@ void mull::SQLiteReporter::reportResults(const std::unique_ptr<Result> &result,
 
   sqlite3_close(database);
 
-  outs() << "Results can be found at '" << databasePath << "'\n";
+  outs() << "Results can be found at '" << databasePath << "'\n"
+         << "Use the following command to generate HTML report:\n"
+         << "REPORT=" << databasePath << " " << "make build\n";
```
AlexDenisov: It doesn't belong here!

googletest/src/gtest-all.cc
```diff
@@ -0,0 +1,48 @@
...
+#include "gtest/gtest.h"
+
+// The following lines pull in the real gtest *.cc files.
+#include "src/gtest.cc"
+#include "src/gtest-death-test.cc"
+#include "src/gtest-filepath.cc"
+#include "src/gtest-port.cc"
+#include "src/gtest-printers.cc"
+#include "src/gtest-test-part.cc"
+#include "src/gtest-typed-test.cc"
```
AlexDenisov: Why do we need gtest in the tree?

lib/SQLiteReporter.cpp
stanislaw: It does belong where?! For now it is a very nice convenience thing, what's your better suggestion?

googletest/src/gtest-all.cc
stanislaw: Note that I have removed a build script that fetches and builds googletest into `.a` files and replaced them with CMake target.

GoogleTest in `.a` files is needed for `MullUnitTests` to run.

GoogleTest in `.bc` files is needed for `mull-unittests-bitcode` to find their GoogleTest symbols when auto-testing.

I find that having GoogleTest in-source is the easiest way to get GoogleTest built and linked in automatic way out of CMake's box.

I am of course open for better suggestions ;)

stanislaw:
@AlexDenisov if these are the only remarks you have, I am happy to consider my work done :trollface: .

stanislaw:
@mbj what you are asking seems to be more advanced level of consideration that I am having here. Most of the time I spent on making a build process to work: complete pipeline that:

1) builds Mull
2) builds Mull's source code as LLVM bitcode
3) 1 tests 2

So I would say that current Mull tests everything of itself as far as what our current three mutation operators can find: AddMutation (`+ -> -`), RemoveVoidCall, NegateCondition.

I need more time of wearing analyst's, not implementor's, hat to speak more about what we are exactly doing here :)

lib/SQLiteReporter.cpp
AlexDenisov: Well, this thing doesn't work on my machine because I use another shell.
Also, `make build` is part of another system.
Please, let's not add it here.

mbj:
@stanislaw Does this mean that the tests mull comes with can run against the LLVM bitcode from 2) to kill mutations applied to it by 1)?

lib/SQLiteReporter.cpp
stanislaw: @AlexDenisov I have to say that this appears to be one of the most useful features. I would rather change the wording but definitely leave it here. It is super convenient to speed up the iteration process.

AlexDenisov: Just copy the database path and then run this command in your shell.
```
make build REPORT=`pbpaste`
```
Next time you would just need to copy the database path and hit `up` button. That's it.

stanislaw:
Closing this PR. 

- Google Test has been merged in a separate PR
- Mull Auto Test now lives in a [separate repository](https://github.com/mull-project/mull-auto-test) that has to be embedded in `projects` directory as of #164.

PR #147, opened 2017-04-22
Config class: introduce bitcodeFileList parameter.

stanislaw:


lib/Driver.cpp
```diff
@@ -273,23 +271,30 @@ std::vector<llvm::object::ObjectFile *> Driver::AllObjectFiles() {
...
-    auto OwnedModule = Loader.loadModuleAtPath(ModulePath);
+  std::vector<std::string> bitcodePaths = Cfg.getBitcodePaths();
+  std::vector<unique_ptr<MullModule>> modules =
+    Loader.loadModulesFromBitcodeFileList(bitcodePaths);
+
+//  for (auto &ownedModule : modules) {
 //    assert(OwnedModule && "Can't load module");
 //    Ctx.addModule(std::move(OwnedModule));
-  }
+//  }
```
AlexDenisov: Let's just get rid of those comments.

stanislaw: Let's do it in a separate scope: https://github.com/mull-project/mull/issues/149.

Issue #115, opened 2017-02-18
Continuous Integration (Travis)


stanislaw:
We are moving to out-of-source hosting so the process will be the following:

- Build on Mac OS image
- `brew install llvm@3.9`
- clone GoogleTest and compile it (maybe better to embed binary to the source tree)
- clone Mull
- `make test_unit` (builds Ninja, then `ninja MullUnitTests`)

Later:

- Install Rust (mainly we need Rust's `libstd` and `libtest`). Can be done using `rustup`.

PR #148, opened 2017-04-22
Add .travis.yml

Quolyk:
#115
Unfortunately, It seems impossible for Travis script to support Rust at the moment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+language: cpp
+
+compiler:
+- clang
+
+os:
+- osx
+
+before_install:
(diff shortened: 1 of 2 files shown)


stanislaw:
@Quolyk thanks for taking time to fight with Travis. The [build log](https://travis-ci.org/Quolyk/mull/builds/224720836) looks good.

A bit more portable and convenient way of writing a CI file would be to encapsulate all required build steps in a custom `Makefile.CI` file so that `.travis.yml` only references `make -f Makefile.CI`.

But I think we can do it later and let's wait for @AlexDenisov's thoughts.

PR #150, opened 2017-04-24
Clean up Driver debugs

Quolyk:
#149


AlexDenisov:
@Quolyk I am not exactly sure what @stanislaw meant in the #149, my version is below 😄

As far as I know, I was the only one who used these methods a long time ago, meaning that we don't need them anymore. So, feel free to just drop them completely. Also, do not forget to remove them from the `driver.cpp` among with command line options here: https://github.com/mull-project/mull/blob/master/tools/driver/driver.cpp#L40

stanislaw:
+1 to what @AlexDenisov just said. I didn't want to handle this in my completely different PR so I opened #149.

Issue #99, opened 2017-02-02
Config: add `dynamic_libraries` option to provide additional libraries for Mull to load.


stanislaw:
Example:

```
fork: true
max_distance: 2
bitcode_files:
- ...
dynamic_libraries:
- /Users/Stanislaw/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/libtest-551a7b69d9c2ff2f.dylib
```

stanislaw:
The second part of the quest:

to validate the paths - print a warning if a path doesn't exists

The third part and we don't know where/how will it be:

Enumerate through the paths and `dlopen` each library. Most likely it is the beginning of `Driver::Run()` method.

stanislaw:
cc @Quolyk

PR #151, opened 2017-04-28
Add dynamicLibrary file check

Quolyk:
#99

### DIFF
--- a/include/Config.h
+++ b/include/Config.h
@@ -34,7 +34,7 @@ class Config {
   std::string testFramework;
 
   std::vector<std::string> mutationOperators;
-  std::vector<std::string> dynamicLibraries;
+  std::string dynamicLibraryFileList;
   std::vector<std::string> tests;
 
   bool fork;
@@ -62,7 +62,7 @@ public:
       //   RemoveVoidFunctionMutationOperator::ID
       // }
     ),
-    dynamicLibraries(),
+    dynamicLibraryFileList(),
     tests(),
     fork(true),
     dryRun(false),
@@ -77,7 +77,7 @@ public:
          const std::string &project,
          const std::string &testFramework,
          const std::vector<std::string> mutationOperators,
-         const std::vector<std::string> libraries,
+         const std::string &dynamicLibraryFileList,
          const std::vector<std::string> tests,
          bool fork,
          bool dryrun,
@@ -89,7 +89,7 @@ public:
     projectName(project),
     testFramework(testFramework),
     mutationOperators(mutationOperators),
-    dynamicLibraries(libraries),
+    dynamicLibraryFileList(dynamicLibraryFileList),
     tests(tests),
     fork(fork),
     dryRun(dryrun),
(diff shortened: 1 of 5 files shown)


include/Config.h
```diff
@@ -29,12 +29,12 @@ namespace mull {
 
 class Config {
   std::string bitcodeFileList;
+  std::string dynamicLibraryFileList;
```
stanislaw: I don't like the fact that you moved `dynamicLibraryFileList` to be the second in order. `bitcodeFileList` goes first because it is the only required parameter in a config.yml. Then, as I see it, parameters more important than dynamicLibraryFileList should go. I would suggest you to move it where it was. But let's also listen to @AlexDenisov .

include/Config.h
```diff
@@ -62,7 +63,6 @@ class Config {
       //   RemoveVoidFunctionMutationOperator::ID
       // }
     ),
-    dynamicLibraries(),
```
stanislaw: The same: I see no reason to make dynamicLibraryFileList be second in order.

include/Config.h
```diff
@@ -167,24 +183,25 @@ class Config {
...
+    if (mutationOperators.empty() == false) {
       Logger::debug() << "\t" << "mutation_operators: " << '\n';
 
-      for (auto mutationOperator : getMutationOperators()) {
+      for (auto mutationOperator : mutationOperators) {
         Logger::debug() << "\t- " << mutationOperator << '\n';
       }
     }
-
-    if (getTests().empty() == false) {
```
stanislaw: For a future: please avoid doing changes that are not directly relevant to a subject of your PRs. Those changes become a noise for reviewers of your code. Let's always try to do separate PRs (smaller - better).

include/Config.h
```diff
@@ -199,16 +216,26 @@ class Config {
       return errors;
     }
 
-    std::ifstream f(bitcodeFileList.c_str());
+    std::ifstream bitcodeFile(bitcodeFileList.c_str());
```
stanislaw: 👍

include/Config.h
```diff
@@ -199,16 +216,26 @@ class Config {
...
       error << "bitcode_file_list parameter points to a non-existing file: "
             << bitcodeFileList;
 
       errors.push_back(error.str());
-      return errors;
+    }
+    
+    std::ifstream dynamicLibraryFile(dynamicLibraryFileList.c_str());
+    
+    if (dynamicLibraryFile.good() == false && dynamicLibraryFileList.empty() == false) {
```
stanislaw: Given `dynamicLibraryFile` is not a required but optional parameter, please do it like:

```cpp
if (dynamicLibraryFileList.empty() == false) {
  std::ifstream dynamicLibraryFile(dynamicLibraryFileList.c_str()); 
  if (dynamicLibraryFile.good() == false) {
    ...
  }
}
```

unittests/ConfigParserTests.cpp
```diff
@@ -43,6 +42,66 @@ TEST_F(ConfigParserTestFixture, loadConfig_bitcodeFileList) {
   ASSERT_EQ(errors.size(), 0U);
 }
 
+TEST_F(ConfigParserTestFixture, loadConfig_dynamicLibraryFileList_nonExistingFile) {
+  static std::string BitcodeFileList = "/tmp/bitcode_file_list.txt";
```
stanislaw: Why `static`? Did you rather mean `const` 😄 ?

stanislaw: Also please always use `camelCase` for local variables: `bitcodeFileList`, not `BitcodeFileList`. Historically, @AlexDenisov started mull to follow LLVM conventions that include capitalization of local variables but later we both couldn't stand it and switched to a `lowerCamelCase`. So currently the code has both styles and the preferred one we are moving to is `lowerCamelCase`. Btw the same goes to class instance variables.

stanislaw: Ok, I see: it was me to introduced it accidentally. It is `const` of course.

stanislaw:
Thanks @Quolyk !

PR #155, opened 2017-05-01
ForkProcessSandbox: better handling of abnormal exit scenarios.

stanislaw:


lib/ForkProcessSandbox.cpp
```diff
@@ -109,6 +110,13 @@ mull::ForkProcessSandbox::run(std::function<void (ExecutionResult *)> function,
         result.RunningTime = duration_cast<std::chrono::milliseconds>(elapsed).count();
         result.Status = Crashed;
         *sharedResult = result;
+      } else if (WIFEXITED(status) &&
+                 (WEXITSTATUS(status) != 0 || sharedResult->Status == Invalid)) {
```
AlexDenisov: Why checking status?

stanislaw: I think we are missing some kind of `AbnormalExit` status. For example, consider GoogleTestRunner [here](https://github.com/mull-project/mull/blob/96e8077b638fc523c67238f208ec45b3ea6296a2/lib/GoogleTest/GoogleTestRunner.cpp#L199): if we reach that code then we either have Passed or Failed. But if for whatever reason we don't reach there then it should be another status like `AbnormalExit` or what we have now: just `Failed`: https://github.com/mull-project/mull/pull/155/files/9cccb886edd400d1cd2c32b36f1f9fb0b1dbf8c3#diff-0b05a421da879026c1ea9182592a8127R118.

AlexDenisov: I mean this one `sharedResult->Status == Invalid`

stanislaw:
@AlexDenisov, does this make more sense to you now? Please see the latest commit.

Issue #159, opened 2017-05-11
Add default mutation operators to Config constructor


Quolyk:
This task includes following. Suppose we init config from yaml with mutaton operators: `add`, `negate`. If Config constructor has default mutation operators : `negate`, `add`, `remove`, than `YAMLTraits` will map: `add` -> `negate`, `negate` -> `add`, thus producing following: `add`, `negate`, `remove` instead of `add`, `negate`.

stanislaw:
@Quolyk I now see that you mean something different, not just parsing default 3 values from a YAML file correctly. I would recommend you to add a unit test to demonstrate a specific behavior you're talking about here.

PR #160, opened 2017-05-11
Add default mutation operators

Quolyk:
#159


stanislaw:
@Quolyk , unfortunately you code doesn't do a job well. This is a test that I wrote that still passes with your code.

We do expect `.size() == 3` there after your changes but it is still `.size == 0`.

So basically your change does nothing.

PR #182, opened 2017-06-07
Numerous fixes

AlexDenisov:


lib/GoogleTest/GoogleTestMutationOperatorFilter.cpp
```diff
@@ -28,6 +28,10 @@ bool GoogleTestMutationOperatorFilter::shouldSkipTesteeFunction(llvm::Function *
     return true;
   }
 
+  if (testee->getName().find(StringRef("clang_call_terminate")) != StringRef::npos) {
```
stanislaw: +1, saw this one many times.

lib/GoogleTest/GoogleTestFinder.cpp
```diff
@@ -109,12 +111,14 @@ std::vector<std::unique_ptr<Test>> GoogleTestFinder::findTests(Context &Ctx) {
...
       /// compare the beginning of the typename
 
-      if (!STy->getName().startswith(StringRef("class.testing::TestInfo"))) {
+      if (!structType->getName().startswith(testInfoTypeName)) {
         continue;
       }
 
-      /// At this point the Global has only one usage
-      /// The user of the Global is part of initialization function
+      /// Normally the globalValue has only one usage, ut LLVM could add
```
stanislaw: typo :trollface:

lib/GoogleTest/GoogleTestFinder.cpp
```diff
@@ -109,12 +111,14 @@ std::vector<std::unique_ptr<Test>> GoogleTestFinder::findTests(Context &Ctx) {
       /// Hence we cannot just compare string, and rather should
       /// compare the beginning of the typename
 
-      if (!STy->getName().startswith(StringRef("class.testing::TestInfo"))) {
```
stanislaw: 👍

lib/GoogleTest/GoogleTestFinder.cpp
```diff
@@ -228,19 +238,6 @@ std::vector<std::unique_ptr<Test>> GoogleTestFinder::findTests(Context &Ctx) {
   return tests;
 }
 
-std::string demangle(const char* name) {
```
stanislaw: 👍 but we might need it for a display name later on in some other class (to display logs for example?).

AlexDenisov: Then we would add it :)

lib/GoogleTest/GoogleTestFinder.cpp
AlexDenisov: @stanislaw open a ticket ;)

stanislaw: Will do thanks for the hint 👍

PR #213, opened 2017-10-06
[WIP][RFC]: IDE Diagnostics

stanislaw:
1) `IMutationPoint` interface is introduced (just picked up the easiest name for the fastest changes everywhere). 
2) `MutationPoint` now conforms to `IMutationPoint`.
3) `NegateConditionMutationPoint` class is introduced to contain information specific for `NegateCondition` mutation.
4) Observe that within `NegateConditionMutationPoint` and `NegateConditionMutationOperator` there are no casts at all. They interact about their specifics only with each other, no specific interaction happens with outside.

### DIFF
--- a/include/GoogleTest/GoogleTestFinder.h
+++ b/include/GoogleTest/GoogleTestFinder.h
@@ -23,8 +23,8 @@ class MutationPoint;
 
 class GoogleTestFinder : public TestFinder {
   llvm::StringMap<llvm::Function *> FunctionRegistry;
-  std::vector<std::unique_ptr<MutationPoint>> MutationPoints;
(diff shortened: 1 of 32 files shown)


stanislaw:
Looking at the following constructor:

```
NegateConditionMutationPoint(
                               NegateConditionMutationOperator *op,
                               MutationPointAddress address,
                               Value *Val,
                               MullModule *m,
                               CmpInst::Predicate originalPredicate,
                               CmpInst::Predicate negatedPredicate) :
```

it seems evident to me that we are missing some container to accommodate these guys in one unit and this unit would be common to all mutation points because that's their baseline:

```
                               MutationPointAddress address,
                               Value *Val,
                               MullModule *m,
```

This would simplify the interface and property declarations.

PR #220, opened 2017-10-10
MutationPointAddress: move related helpers.

stanislaw:


stanislaw:
Sidenote: `MutationPointAddress` can eventually become `InstructionAddress` (more precise).

include/MutationPoint.h
```diff
@@ -1,25 +1,29 @@
...
 namespace llvm {
 
+class Function;
+class Instruction;
 class Value;
 class Module;
 
 }
 
+using namespace llvm;
```
AlexDenisov: Please don't do this in headers.

lib/IDEDiagnostics.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include <llvm/IR/DebugInfoMetadata.h>
 #include <llvm/IR/Instruction.h>
+#include "llvm/Support/raw_ostream.h"
```
AlexDenisov: Let's use angle brackets.

PR #222, opened 2017-10-16
ScalarValueMutationOperator: diagnostics "from -> to".

stanislaw:
With this PR ScalarValueMutationOperator now outputs the diagnostics in the following format:

```
Scalar Value Replacement: 0 -> 1
```

As discussed I am doing this without a special class for meta information so there is a work that is done twice.

### DIFF
--- a/lib/MutationOperators/ScalarValueMutationOperator.cpp
+++ b/lib/MutationOperators/ScalarValueMutationOperator.cpp
@@ -6,11 +6,11 @@
 #include "MutationPoint.h"
 
 #include <llvm/IR/Constants.h>
-#include "llvm/IR/InstIterator.h"
-#include "llvm/IR/Instructions.h"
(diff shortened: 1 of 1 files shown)


lib/MutationOperators/ScalarValueMutationOperator.cpp
```diff
@@ -149,9 +152,9 @@ static ConstantInt *getReplacementInt(ConstantInt *constantInt) {
   // TODO: Review the rules for mutation.
   // TODO: Didn't find a better way of testing APInt for zero-ness.
   if (intValue.isNegative() == false && intValue.isStrictlyPositive() == false) {
-    replacementIntValue = APInt(intValue.getBitWidth(), 1);
+    replacementIntValue = APInt(8U, 1);
```
AlexDenisov: ?

Issue #234, opened 2017-10-19
 On-disk cache broken


AlexDenisov:
On-disk cache is broken at the moment. The explanation is following.

Recently, we have introduced dynamic call tree: to build the call tree we inject callbacks in every function. To avoid static variables, we call the callback with a pointer to an object that holds the data necessary to build the call tree. Some code to illustrate what's going on:

```
void mull_enterFunction(Driver *driver, uint64_t functionID) { ... }
/// ...
llvm::Function *function; /// some function
uint64_t functionID; /// some ID
Driver *pointer = this; /// pointer to the driver
injectCallbackWithPointerToDriver(function, functionID, pointer);
```

After injection, the code is being compiled and saved on-disk. The problem is that it stores the value of a `pointer` as it was during the execution. When the program runs next time, then nothing is injected and the pointer being reused. Apparently, the actual address of the driver is different which leads to crash.

AlexDenisov:
Fixed via #305

PR #235, opened 2017-10-19
Disable On Disk Cache

AlexDenisov:


AlexDenisov:
Related to https://github.com/mull-project/mull/issues/234

PR #262, opened 2017-10-27
Fix off-by-one error

AlexDenisov:


AlexDenisov:
@stanislaw shall we consider running tests on CI with some sanitizers enabled?..

PR #270, opened 2017-10-31
Use plain files instead of pipes to capture child process output

AlexDenisov:


lib/ForkProcessSandbox.cpp
```diff
@@ -35,23 +35,38 @@ pid_t mullFork(const char *processName) {
...
   }
-
-  if (pipe(stderr_file_descriptor) == -1) {
-    perror("stderr_pipe");
-    exit(1);
+  if (unlink(filename) == -1) {
+    perror("unlink");
   }
 
+  fseek(file, 0, SEEK_END);
```
stanislaw: Let me get this right: you first delete the file and then read from it? Is this a normal practice?

stanislaw: Your method should be smth like: `unlinkFileAndRead` 😄

AlexDenisov: The man page would explain it better than I can
https://linux.die.net/man/2/unlink

stanislaw: Good.

PR #271, opened 2017-10-31
Use one process per test run (instead of 3)

AlexDenisov:


unittests/ForkProcessSandboxTest.cpp
```diff
@@ -66,3 +66,23 @@ TEST(ForkProcessSandbox, statusAbnormalExit_IfExitingWithZeroButResultWasNotSet)
 
   ASSERT_EQ(result.Status, AbnormalExit);
 }
+
+TEST(ForkProcessSandbox, statusTimeout) {
+  ForkProcessSandbox sandbox;
+
+  ExecutionResult result = sandbox.run([&](ExecutionResult *_) {
+    sleep(3);
+  }, Timeout);
```
stanislaw: 👍

PR #281, opened 2017-11-10
Clone module

AlexDenisov:
Everything seems to be working now, but let's not merge it yet.

 - I want to make sure it does not break things
 - I want to measure if memory consumption has dropped
 - I want to clean up the driver a bit, it becomes messy

P.S. Dependent on #277

### DIFF
--- a/include/MullModule.h
+++ b/include/MullModule.h
@@ -2,19 +2,25 @@
 
 #include <string>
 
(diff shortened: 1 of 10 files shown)


AlexDenisov:
For OpenSSL memory consumption dropped from ~200MB to ~170MB.
Everything is working.
Code of the driver is still a mess :(

PR #289, opened 2017-11-21
Add ability to use precompiled object files

AlexDenisov:
It is useful when you have some code compiled from an assembly and there is no way to get bitcode.
Should go after #288.

Blocked by the issue with Orc JIT: https://bugs.llvm.org/show_bug.cgi?id=35371


AlexDenisov:
It works perfect on Linux, but not on MacOS.
I suggest merging it into master.
I assume it may it may work on MacOS as well (under some circumstances), but it is not recommended at the moment.

PR #318, opened 2018-02-06
Cache instrumented object files

AlexDenisov:


AlexDenisov:
@stanislaw good to go.

PR #319, opened 2018-02-08
Optimize sqlite reporter

AlexDenisov:
Goes after #318.
Metrics before/after this PR:

```
FMT:
Saving result: .................... 115ms
Saving result: .................... 10ms

LLVM:
Saving result: .................... 1188ms
Saving result: .................... 10ms
```


AlexDenisov:
Based on hints from this SO thread:
https://stackoverflow.com/questions/1711631/improve-insert-per-second-performance-of-sqlite

PR #333, opened 2018-03-03
Docs/EquivalentMutations: Conditionals Boundary example.

stanislaw:


joakim-brannstrom:
(drive by comment out of interest of the subject) :)

Consideration 1
May be worthwhile to consider how the type information can be used to reduce equivalent mutations for the ROR mutation operator.
In dextool the implementation of type aware schemas for bool/float/enum/pointer reduce equivalent mutants by 10-40%. Note though it is highly dependent on the design of the SUT. Flight critical code loves enums. Googletest? Not so much.
Our conclusion where that ROR is _not_ recommended for large projects if it is not _type aware_ because there are too many junk/equivalent mutants which makes it hard/too expensive if the process requires all survived mutants to be manually inspected.
[dextool ROR](https://github.com/joakim-brannstrom/dextool/blob/master/plugin/mutate/doc/design/mutations.md#relational-operator-replacement-ror)

Consideration 2
Is ROR implemented? If so, is RORG implemented? If not I would strongly recommend it. It reduced the amount of mutants by a considerable amount.
I think I have the numbers laying around somewhere just not in my not-so-elephant-memory.

PR #336, opened 2018-03-10
Fix broken "How to get LLVM Bitcode" link

tomj:
Thanks for a great project!


AlexDenisov:
Thanks @tomj. It still renders an old link for me. Could you please confirm if it works for you?

PR #351, opened 2018-05-21
[WIP] Autogenerate fixtures

AlexDenisov:


AlexDenisov:
This one is getting too complicated, much simpler option would be to use the lab as is, but provide compiler based on current LLVM_ROOT: #353

PR #374, opened 2018-08-16
Thread safe test runners

AlexDenisov:
Goes right after #373


AlexDenisov:
Actually, it is independent from #373

PR #385, opened 2018-09-04
Add functional include

def-:
Fixes compilation with clang 6.0.1 on my machine since std::functional is defined in functional header.


AlexDenisov:
Awesome, thank you @def-.

PR #388, opened 2018-09-08
Add documentation for config file

AlexDenisov:
Let's consider it as a first draft


Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+real-life examples at the end of the document.
+
+### Configuration options
+
+---
+```
+project_name: string
+```
+Project name parameter is used by `SQLiteReporter` to compose a name of an
+outputting SQLite file.
```
stanislaw: output?

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+ - `GoogleTest`: default value. Works out of the box except for typed tests (see #15).
+ - `CustomTest`: used for project with custom test framework. See `custom_tests` below.
+ - `SimpleTest`: used for testing only.
+
+---
+```
+bitcode_file_list: path (string)
+```
+This is the most important part of the configuration. It points to a plain text
+file that contains new-line separated list of bitcode files which Mull has to process.
```
stanislaw: a new-line

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
+# Configuration
+
+Mull uses [yaml](http://yaml.org) format to control its behavior.
+
+Below you can find configuration options and their explanation along with
+real-life examples at the end of the document.
+
```
stanislaw: the real-world?

Docs/Config.md
stanislaw: also can be "newline"

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+---
+```
+test_framework: string
+```
+Tells Mull which test framework adapter to use.
+
+Supported test frameworks:
+
+ - `GoogleTest`: default value. Works out of the box except for typed tests (see #15).
+ - `CustomTest`: used for project with custom test framework. See `custom_tests` below.
```
stanislaw: a project, with a custom

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+This is the most important part of the configuration. It points to a plain text
+file that contains new-line separated list of bitcode files which Mull has to process.
+
+---
+```
+dynamic_library_file_list: path (string)
+```
+
+Similar to the `bitcode_file_list`, but the file contains new-line separated
+list of dynamic libraries(`*.dylib`, `*.so`) required for execution of a
```
stanislaw: an execution

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+ - `math_add_mutator`: replaces `+` with `-`
+ - `math_sub_mutator`: replaces `-` with `+`
+ - `math_div_mutator`: replaces `/` with `*`
+ - `math_mul_mutator`: replaces `*` with `/`
+ - `conditional`: group, includes 3 next mutators
+ - `and_or_replacement_mutator`: replaces `&&` with `||` and vice versa
+ - `negate_mutator`: negates all conditions, e.g.: `!=` -> `=`, `<` -> `>=`
+ - `conditionals_boundary_mutator`: mutates boundaries, e.g.: `<` -> `<=`, `<=` -> `<`
+ - `functions`: group, includes 2 next mutators
+ - `replace_call_mutator`: replaces call to a function returning integer with `42`
```
stanislaw: a call

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+ - `remove_void_function_mutator`: removes call to a function returning `void`
```
stanislaw: a call

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+ - `constant`: group, includes only 1 next mutator
+ - `scalar_value_mutator`: replaces constants with other values: zero -> 1,
```
stanislaw: the constants

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+Lists names of tests that should be taken into account. All the tests that are
+not in this list will be ignored.
+
+---
+```
+exclude_locations: array of strings
+```
+Skips mutation based on its on-disk location.
```
stanislaw: location on the disk?

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+
+---
+```
+dry_run: boolean
+```
+
+Possible values: `true`/`enabled`, `false`/`disabled`. Defaults to `false`.
+
+When enabled Mull will find mutations, but won't execute them.
```
stanislaw: will not execute?

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+
+---
+```
+fail_fast: boolean
+```
+Possible values: `true`/`enabled`, `false`/`disabled`. Defaults to `false`.
+
+A mutant may be accessible from many tests. Normally, Mull executes a mutant
+against each test separately. When `fail_fast` option is enabled Mull stops
+running tests against a mutant as soon as on the tests fail.
```
stanislaw: one of the tests fails

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+
+---
+```
+use_cache: boolean
+```
+Possible values: `true`/`yes`, `false`/`no`. Defaults to `false`.
+
+Saves compiled object files on-disk to reuse on next runs.
```
stanislaw: on disk? (without -)

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+  enabled: boolean
+  cxx_compdb_dir: path (string)
+  cxx_compilation_flags: string
+```
+
+Some mutations found at the Bitcode level cannot be reproduced by a programmer
+at the source code level. Mull comes with `CXXJunkDetector` that filters out
+some junk mutations.
+
+`CXXJunkDetector` is based on `libclang`, in order to parse source files
```
stanislaw: . In order

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+correctly `libclang` needs either compilation database, or set of flags used
```
stanislaw: a compilation database, or a set of

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+for a compilation.
+
+By default junk detection is disabled. See examples below.
```
stanislaw: Junk detection is disabled by default.

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+  workers: integer
+  test_execution_workers: integer
+  mutant_execution_workers: integer
+```
+
+Mull can run most of the tasks in parallel. It does so by default.
+Tests can be run sequentially if they are not designed to be run in parallel.
+
+When running tests Mull reuses one JIT stack across several threads. Though,
+when executing mutants it creates a JIT stack per-thread, which may consume
```
stanislaw: per thread

stanislaw: and this may consume

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+significant amount of RAM. In this case, it may make sense to use less threads
```
stanislaw: fewer threads

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+```
+custom_tests:
+  - name: string
+    method: method name
+    program: program name
+    arguments: string
+```
+
+This option is used with `CustomTest`. If a project uses custom test suite,
+then one can provide list of tests manually. Each test entry should have
```
stanislaw: a list

Docs/Config.md
```diff
@@ -0,0 +1,285 @@
...
+a unique name, name of a program (`argv[0]`), arguments (`argv`) required to
+run the test program, and the name of a test method. The method name should
+contain a mangled name of a test function.
+
+If `CustomTest` used and no `custom_tests` specified, then Mull picks the whole
```
stanislaw: is used

stanislaw: are specified

AlexDenisov:
Should be good enough by now.

PR #394, opened 2018-09-16
[WIP] Add LLVM-3.8 support

AlexDenisov:


AlexDenisov:
LLVM-3.8's `cl` module is missing some parts we use within `mull-reporter`. To add LLVM-3.8 support we would have to use some other solution for command-line argument parsing.

PR #400, opened 2018-10-06
Use i8 instead of void for trampoline pointers

valpackett:
Creating void pointers is forbidden by [LLVM assertions ("Invalid type for pointer element!")](https://llvm.org/doxygen/Type_8cpp_source.html#l00644).
Clang uses pointers to 8-bit integers to represent void pointers, so let's do the same.

Why minus one `->getPointerTo()`? For some reason `injectInstrumentationInfoPointer` was generating a triple pointer, which tripped a different assertion ("Calling a function with a bad signature!") — I have no idea how that happens o_0


AlexDenisov:
Hi @myfreeweb, thanks a lot for the contribution. I must say the solution is a bit hacky and I will need to revise it, which is on my todo list for the project. I will do some testing and then merge the PR.

PR #403, opened 2018-10-13
Use PkgConfig (optional) to find SQLite library and include dirs

AlexDenisov:


AlexDenisov:
Just checked it on FreeBSD and it works.

PR #410, opened 2018-10-30
MullModule: CallInst name has to be empty for void functions

myfreeweb:
Fixes assertion failure:

```
Assertion failed: (!getType()->isVoidTy() && "Cannot assign a name to void values!"), function setNameImpl, file /wrkdirs/usr/ports/devel/llvm60/work/llvm-6.0.1.src/lib/IR/Value.cpp, line 247.
```

http://lists.llvm.org/pipermail/llvm-dev/2016-March/096242.html


myfreeweb:
Do you test with LLVM 6 with assertions enabled?

AlexDenisov:
Ohh, no, we do work against precompiled versions most of the time. This is a good point, I will include it into our CI as well.

myfreeweb:
You can do it with your own compilation of LLVM, just make sure to set `LLVM_ENABLE_ASSERTIONS`

AlexDenisov:
@myfreeweb thank you for the hint, I will do that soon.

PR #433, opened 2019-02-08
Add a test case involving exceptions

AlexDenisov:


AlexDenisov:
One test in isolation works, but when run all together then the test with exceptions fails.
Putting on hold for a time being.

PR #499, opened 2019-04-14
CPack: do not error on unknown OS; mull-cxx: support shared libLLVM

myfreeweb:
Fixes the following error:

```
CMake Error at cmake/cpack.cmake:47 (if):
  if given arguments:

    "STREQUAL" "macOS"

  Unknown arguments specified
```

### DIFF
--- a/cmake/cpack.cmake
+++ b/cmake/cpack.cmake
@@ -41,6 +41,9 @@ else()
       endif()
     endforeach()
 
+  else()
+    set (CPACK_SYSTEM_NAME "unknown")
+
   endif()
 endif()
 
--- a/tools/driver-cxx/CMakeLists.txt
+++ b/tools/driver-cxx/CMakeLists.txt
@@ -3,8 +3,18 @@ set (SOURCES
   ${CMAKE_CURRENT_LIST_DIR}/DynamicLibraries.cpp
 )
 
+if ("LLVM" IN_LIST LLVM_AVAILABLE_LIBS)
+  set (MULL_CXX_LLVM_LIBRARIES
+    LLVM
+  )
+else()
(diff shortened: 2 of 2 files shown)


AlexDenisov:
FreeBSD, I suppose? :)

myfreeweb:
Yes

AlexDenisov:
Thanks a lot for the fix, I'll merge on green.
jfyi, there is an (empty) issue for the FreeBSD package #449.

myfreeweb:
Also fixed an over-linking issue: `mull-cxx` would get linked to both `libLLVM-8.so` and various static libs (`libLLVMCore.a` etc.) which would cause it not to work:

```
: CommandLine Error: Option 'print-summary-global-ids' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```

(https://github.com/AlexDenisov/LibEBC/pull/2 is part of that too)

myfreeweb:
(btw, `llvm_config(target USE_SHARED Core Support Whatever…)` automatically handles shared vs static, but I don't know if you have any reasons to do it manually and I don't know if it's available on old LLVMs)

AlexDenisov:
Good catch.

> (btw, llvm_config(target USE_SHARED Core Support Whatever…) automatically handles shared vs static, but I don't know if you have any reasons to do it manually and I don't know if it's available on old LLVMs)

Does it use `llvm-config` binary under the hood? We had some troubles with it in the past, e.g. linking order of clang libraries. Other than that - current build system is something that somehow works, it was not properly engineered, but rather evolved with time.

myfreeweb:
No, it's a small wrapper around `llvm_map_components_to_libnames`. Nothing in `lib/cmake` actually seems to call `llvm-config`.

PR #517, opened 2019-05-19
Reporting JSON to Mutation Testing Elements 

stanislaw:


include/mull/JunkDetection/CXX/Visitors/InstructionRangeVisitor.h
```diff
@@ -15,11 +15,12 @@ class InstructionRangeVisitor {
   InstructionRangeVisitor(const VisitorParameters &parameters);
   void visitRangeWithLocation(const clang::SourceRange &range);
   bool foundRange();
-
+  clang::SourceRange getSourceRange();
 private:
   const clang::SourceManager &sourceManager;
   const clang::SourceLocation &sourceLocation;
   clang::SourceRange sourceRange;
+  clang::SourceRange __sourceRange;
```
AlexDenisov: Why two variables? Some artefact?

include/mull/MutationResult.h
```diff
@@ -22,6 +22,9 @@ class MutationResult {
   MutationPoint *getMutationPoint() { return MutPoint; }
   int getMutationDistance() { return distance; }
   Test *getTest() { return test; }
+  bool hasMutantSurvived() const {
+    return Result.status == ExecutionStatus::Passed;
+  }
```
AlexDenisov: I'd say the `survived` is not the property that should exists on the level of `libmull`. There are at least two different ways to decide whether the mutant is survived, or not. I'd put this into a reporter.

include/mull/Reporters/MutationTestingElementsReporter.h
```diff
@@ -0,0 +1,22 @@
...
+#include "mull/MutationPoint.h"
+
+#include <clang/Basic/SourceLocation.h>
+
+#include <map>
+
+namespace mull {
+class MutationPoint;
+class MutationTestingElementsReporter : public Reporter {
+  const std::map<MutationPoint *, BeginEndPair> &mutationPointsRanges;
```
AlexDenisov: Shouldn't a mutation point have two pairs? I.e. 'begin line/column' and 'end line/column'

lib/CMakeLists.txt
```diff
@@ -84,7 +84,17 @@ set(mull_sources
...
+  JunkDetection/CXX/Visitors/MathMulVisitor.cpp
+  JunkDetection/CXX/Visitors/MathDivVisitor.cpp
+  JunkDetection/CXX/Visitors/ReplaceCallVisitor.cpp
+
+  Reporters/IDEReporter.cpp
+  Reporters/MutationTestingElementsReporter.cpp
+  Reporters/SourceManager.cpp
+
+  TestFrameworks/NativeTestRunner.cpp
+)
```
AlexDenisov: Nice one, I didn't pay attention to this file, seems like CLion could do a better job here.

lib/Config/RawConfig.cpp
```diff
@@ -355,6 +367,21 @@ std::vector<std::string> RawConfig::validate() {
     }
   }
 
+  if (!mutationTestingElementsReportPath.empty()) {
+    std::string reportPathDir = getFileDir(mutationTestingElementsReportPath);
+
+    if (!llvm::sys::fs::exists(reportPathDir)) {
```
AlexDenisov: We could try creating this directory rather than failing immediately.

lib/JunkDetection/CXX/CXXJunkDetector.cpp
```diff
@@ -17,26 +17,45 @@
...
   if (ast->isInSystemHeader(location)) {
     return true;
   }
 
-  VisitorParameters parameters = {.sourceManager = ast->getSourceManager(),
-                                  .sourceLocation = location,
-                                  .astContext = ast->getASTContext()};
+  VisitorParameters parameters = {.sourceManager = sourceManager,
+    .sourceLocation = location,
+    .astContext = ast->getASTContext()};
```
AlexDenisov: Do you use clang-format?

lib/JunkDetection/CXX/CXXJunkDetector.cpp
```diff
@@ -17,26 +17,45 @@
 using namespace mull;
 
 template <typename Visitor>
-static bool isJunkMutation(ASTStorage &storage, MutationPoint *point) {
+static bool
+isJunkMutation(ASTStorage &storage, MutationPoint *point,
+               std::map<MutationPoint *, BeginEndPair> &mutationPointsRanges) {
```
AlexDenisov: i'm curious if there is a better way to store/populate this information.
It would be nice if one can use the ASTStorage from a reporter and having access to the AST would be very helpful.

For a pretty long time I'm thinking how to implement something like `DiffReporter` that will report mutants as git patches.

lib/JunkDetection/CXX/CXXJunkDetector.cpp
```diff
@@ -17,26 +17,45 @@
...
+    clang::SourceRange sourceRange = visitor.getSourceRange();
+    clang::SourceLocation sourceLocationBegin = sourceRange.getBegin();
+    clang::SourceLocation sourceLocationEnd = sourceRange.getEnd();
+
+    unsigned int beginCol =
+      sourceManager.getExpansionColumnNumber(sourceLocationBegin);
+    unsigned int endCol =
+      sourceManager.getExpansionColumnNumber(sourceLocationEnd);
+
+    mutationPointsRanges[point] = std::make_pair(beginCol, endCol);
```
AlexDenisov: As mentioned earlier, this is not enough: what happens when an AST node is scattered across different lines?

include/mull/Reporters/SourceManager.h
```diff
@@ -0,0 +1,30 @@
...
+
+  ~SourceManager();
+
+private:
+  std::map<std::string, LineOffset> lineOffsets;
+
+  LineOffset &getLineOffset(const SourceLocation &location);
+};
+
+} // namespace mull
```
AlexDenisov: Would you mind submitting this as a separate pull request? It makes sense to extract it to an interface.

lib/Config/RawConfig.cpp
```diff
@@ -14,6 +14,12 @@ int MullDefaultTimeoutMilliseconds = 3000;
 
 using namespace mull;
 
+static std::string getFileDir(const std::string &filePath) {
+  std::size_t found = filePath.find_last_of("/\\");
+  std::string fileDir = filePath.substr(0, found);
+  return fileDir;
+}
```
AlexDenisov: Would it make sense to use `dirname` (`man 3 dirname`)? There could also be such functions in the `llvm::fs` or `llvm::path` namespace.

lib/Reporters/MutationTestingElementsReporter.cpp
```diff
@@ -0,0 +1,154 @@
+#include "mull/Reporters/MutationTestingElementsReporter.h"
+
+#include "mull/Config/RawConfig.h"
+#include "mull/Logger.h"
+#include "mull/MutationResult.h"
+#include "mull/Mutators/Mutator.h"
+#include "mull/Reporters/SourceManager.h"
+#include "mull/Result.h"
+
+#include "json11/json11.hpp"
```
AlexDenisov: Let's use angle brackets for third-parties.

tests/MutationTestingElementsReporterTest.cpp
```diff
@@ -0,0 +1,126 @@
...
+                                                mutatedTestExecutionResult};
+
+  std::ifstream t("/tmp/mull.mutation-testing-elements.json");
+  std::string str((std::istreambuf_iterator<char>(t)),
+                  std::istreambuf_iterator<char>());
+  std::string err;
+  Json object = Json::parse(str, err);
+
+  std::cout << object.dump() << '\n';
+  ASSERT_EQ(object.is_null(), false);
```
AlexDenisov: `ASSERT_FALSE`?

include/mull/Config/ConfigParser.h
```diff
@@ -149,6 +149,8 @@ template <> struct MappingTraits<mull::RawConfig> {
     io.mapOptional("cache_directory", config.cacheDirectory);
     io.mapOptional("junk_detection", config.junkDetection);
     io.mapOptional("parallelization", config.parallelizationConfig);
+    io.mapOptional("mutation_testing_elements_report_path",
+                   config.mutationTestingElementsReportPath);
```
AlexDenisov: You would also need to add this to the `mull-cxx`.

stanislaw: Could you explain what you are referring to?

include/mull/Reporters/MutationTestingElementsReporter.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include "Reporter.h"
+
+#include "mull/MutationPoint.h"
+#include "mull/JunkDetection/CXX/ASTStorage.h"
```
AlexDenisov: These two can be forward-declared (`class ASTStorage;`)

include/mull/JunkDetection/CXX/ASTStorage.h
```diff
@@ -12,6 +12,13 @@ namespace mull {
 
 class MutationPoint;
 
+struct MutationPointSourceInfo {
+  int beginLine;
+  int beginColumn;
+  int endLine;
+  int endColumn;
+};
+
```
AlexDenisov: I don't think it belongs here. I'd suggest putting it into the reporter.
Do you see a different use-case which requires this struct to be defined here?

include/mull/JunkDetection/CXX/ASTStorage.h
```diff
@@ -37,11 +44,17 @@ class ASTStorage {
 
   ThreadSafeASTUnit *findAST(const MutationPoint *point);
 
+  MutationPointSourceInfo getSourceInfo(MutationPoint *mutationPoint);
+
+  void setSourceRange(MutationPoint *mutationPoint,
+                      clang::SourceRange sourceRange);
```
AlexDenisov: I think it's better and probably more flexible to store an AST node here instead of the range.
Do you foresee any issues with this approach?

include/mull/JunkDetection/CXX/Visitors/AndOrReplacementVisitor.h
```diff
@@ -13,6 +13,7 @@ class AndOrReplacementVisitor
   AndOrReplacementVisitor(const VisitorParameters &parameters);
   bool VisitBinaryOperator(clang::BinaryOperator *binaryOperator);
   bool foundMutant();
+  clang::SourceRange getSourceRange();
```
AlexDenisov: As mentioned above: pointer to a node should be sufficient.

lib/Config/RawConfig.cpp
```diff
@@ -14,6 +16,14 @@ int MullDefaultTimeoutMilliseconds = 3000;
 
 using namespace mull;
 
+static std::string getFileDir(const std::string &filePath) {
+  assert(filePath.size() < 256 &&
+         "Expected file path string to be of realistic size");
+  char path[256];
+  strncpy(path, filePath.c_str(), filePath.size() + 1); // + 1 goes for '\0'
+  return std::string(path);
+}
```
AlexDenisov: Something is missing here, right?

lib/JunkDetection/CXX/ASTStorage.cpp
```diff
@@ -136,3 +136,47 @@ ThreadSafeASTUnit *ASTStorage::findAST(const MutationPoint *point) {
...
+      sourceManager.getExpansionColumnNumber(sourceLocationBegin);
+  info.beginLine =
+      sourceManager.getExpansionLineNumber(sourceLocationBegin, nullptr);
+  info.endColumn =
+      sourceManager.getExpansionColumnNumber(sourceLocationEndActual);
+  info.endLine =
+      sourceManager.getExpansionLineNumber(sourceLocationEndActual, nullptr);
+
+  return info;
+}
```
AlexDenisov: As mentioned earlier, it does not belong here. Unless there is a strong reason that I'm missing.

tools/driver-cxx/driver-cxx.cpp
```diff
@@ -292,6 +293,10 @@ int main(int argc, char **argv) {
 
   mull::IDEReporter ideReporter;
   ideReporter.reportResults(*result, rawConfig, metrics);
+
+  mull::MutationTestingElementsReporter mutationTestingElementsReporter(junkDetector.getMutationPointsRanges());
```
AlexDenisov: Does it compile? I cannot find this method, and the reporter takes the ASTStorage now.

include/mull/Config/ConfigParser.h
AlexDenisov: I don't see how to configure it via `mull-cxx`.

AlexDenisov:
Forgot to mention: is it possible to parallelize this reporter?

lib/Config/RawConfig.cpp
stanislaw: 👍

stanislaw: I am not yet testing those paths, only the default fallback to `/tmp/....json` file.

stanislaw: Actually, do you think it is a good idea to extract this function to some FileUtils and write a unit test for it?

P.S. I could not find anything in `llvm::sys::fs`.

stanislaw: `mull::file_utils`?

AlexDenisov: I have very strong objections against `*Utils` :)

stanislaw: Then just follow the approach of llvm:

`mull::fs` or `mull::sys::fs`...

The point is to have it in a dedicated place with some other helpers that we had. Also with some tests.

AlexDenisov: Do we have other helpers? I don't mind having a separate place, but not with one function

AlexDenisov: You can look into the latest version of LLVM, if there is such function, then we can backport it under `llvm_compat` namespace.

lib/JunkDetection/CXX/ASTStorage.cpp
stanislaw: This will change, let's agree on setting up the AST nodes first.

lib/Config/RawConfig.cpp
stanislaw: I found nothing relevant in the latest llvm::sys::fs docs.

lib/JunkDetection/CXX/ASTStorage.cpp
```diff
@@ -136,3 +121,18 @@ ThreadSafeASTUnit *ASTStorage::findAST(const MutationPoint *point) {
   astUnits[sourceFile] = std::unique_ptr<ThreadSafeASTUnit>(threadSafeAST);
   return threadSafeAST;
 }
+
+clang::Expr *ASTStorage::getMutantASTNode(MutationPoint *mutationPoint) {
+  if (mutantASTNodes.count(mutationPoint) == 0) {
```
AlexDenisov: lock is missing

include/mull/Reporters/ASTSourceInfoProvider.h
```diff
@@ -0,0 +1,28 @@
...
+
+class SourceInfoProvider {
+public:
+  virtual MutationPointSourceInfo getSourceInfo(MutationPoint *mutationPoint) = 0;
+};
+
+class ASTSourceInfoProvider: public SourceInfoProvider {
+  ASTStorage &astStorage;
+public:
+  ~ASTSourceInfoProvider() {}
```
AlexDenisov: Please, use `= default` instead. It is more readable, imo.

include/mull/Reporters/MutationTestingElementsReporter.h
```diff
@@ -0,0 +1,23 @@
...
+class MutationTestingElementsReporter : public Reporter {
+  std::string outputPath;
+  SourceInfoProvider &sourceInfoProvider;
+
+public:
+  MutationTestingElementsReporter(std::string outputPath,
+                                  SourceInfoProvider &sourceInfoProvider);
+  void reportResults(const Result &result, const RawConfig &config,
+                     const Metrics &metrics) override;
+};
```
AlexDenisov: I don't think we agreed on this explicitly, but so far the style is to put public methods first, and then private methods, i.e.:
```
class MutationTestingElementsReporter : public Reporter {
public:
  MutationTestingElementsReporter(...);
  void reportResults(...) override;
private:
  std::string outputPath;
  SourceInfoProvider &sourceInfoProvider;
};
```

tools/driver-cxx/driver-cxx.cpp
```diff
@@ -100,6 +102,18 @@ llvm::cl::opt<std::string> CacheDir(
     llvm::cl::desc("Where to store cache (defaults to /tmp/mull-cache)"),
     llvm::cl::cat(MullCXXCategory), llvm::cl::init("/tmp/mull-cache"));
 
+llvm::cl::opt<bool> ElementsReporterEnabled(
+    "reporter-elements", llvm::cl::Optional,
```
AlexDenisov: Ideally, we should provide an option to enable several reporters, the same way as we do with mutators.

But we can do it separately.

stanislaw: let's do it separately

lib/JunkDetection/CXX/ASTStorage.cpp
stanislaw: omg 😲

README.md
```diff
@@ -91,6 +91,35 @@ list here:
...
+**Note:** Reporting to Mutation Testing Elements requires presence of the 
+`-compdb-path` and `-compilation-flags` (see Junk Mutations above) because Mull 
+needs the compilation database/flags information to reason about the source code 
+in which the mutations are found.
+
+```
+mull-cxx --help
+...
+  -elements-output-path=<string>   - Where to store the Mutation Testing Elements JSON report (defaults to /tmp/mull.mutation-testing-elements.json)
+  -reporter-elements               - Enables Mutation Testing Elements Reporter
```
AlexDenisov: I suggest using the same approach as with mutator, i.e.: `-reporters=sqlite -reporters=elemets`.

But it's fine, I can do it on the next iteration.

include/mull/Mutators/AndOrReplacementMutator.h
```diff
@@ -32,7 +32,9 @@ class MutationPointAddress;
 
 class AndOrReplacementMutator : public Mutator {
 
-  AND_OR_MutationType
+  AND_OR_MutationType findPossibleMutation(Value &V);
+
+    AND_OR_MutationType
   findPossibleMutationInBranch(BranchInst *branchInst,
```
AlexDenisov: Seems like formatting is not correct.

Did you format the whole patch?

README.md
stanislaw: Let's do it on the next iteration, please 🙏.

include/mull/Mutators/AndOrReplacementMutator.h
stanislaw: Yes, I did. This one was from my cat 😸

AlexDenisov: Weird, GitHub renders line 37 with indentation of 4 spaces and line 38 with 2 spaces.

README.md
AlexDenisov: Deal.

PR #526, opened 2019-06-06
Fix some spelling mistakes

srasku:
These appear to be mostly typos.


AlexDenisov:
Good catch, thanks a lot @srasku!

PR #532, opened 2019-06-16
WIP: Possible Bug Report: the LLVM Modules and Instructions content gets corrupted before mull finishes execution

stanislaw:
In order to proceed with finishing the work on Mutation Testing Elements support,
we need to make sure that MutationTestingElementsReporter can have access to
original LLVM Module/Instructions content that is used for generating mutation
points.

This small changeset on top of the latest master branch shows that by the
time we want to call the MutationTestingElements to generate the JSON report,
the LLVM Module/Instruction content gets corrupted.

The example demonstrates that the critical asserts of this changeset are hit
instead of going further. It is important that at the moment the Junk Detection
is run, the same mutation points works inside findAST() method, but then
it hits the assert when the findAST() is called later in the end of the driver-cxx.

This is a minimal example that reproduces the issue found during this work:

WIP: Reporting JSON to Mutation Testing Elements #517,
https://github.com/mull-project/mull/pull/517

[skip ci]

### DIFF
--- a/include/mull/JunkDetection/CXX/CXXJunkDetector.h
+++ b/include/mull/JunkDetection/CXX/CXXJunkDetector.h
@@ -10,13 +10,13 @@ struct JunkDetectionConfig;
 
 class CXXJunkDetector : public JunkDetector {
 public:
-  explicit CXXJunkDetector(JunkDetectionConfig &config);
+  explicit CXXJunkDetector(ASTStorage &astStorage);
   ~CXXJunkDetector() override = default;
 
   bool isJunk(MutationPoint *point) override;
 
 private:
-  ASTStorage astStorage;
+  ASTStorage &astStorage;
 };
 
 } // namespace mull
--- a/lib/JunkDetection/CXX/ASTStorage.cpp
+++ b/lib/JunkDetection/CXX/ASTStorage.cpp
@@ -96,7 +96,13 @@ ThreadSafeASTUnit *ASTStorage::findAST(const MutationPoint *point) {
   assert(point);
   assert(!point->getSourceLocation().isNull() && "Missing debug information?");
 
+  errs() << "mutation point: " << point << "\n";
+
   auto instruction = dyn_cast<Instruction>(point->getOriginalValue());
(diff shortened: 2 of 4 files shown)


stanislaw:
Also notice, that sometimes it passes the `assert(instruction)` assert and crashes inside `instruction->getModule`.

This is the command I am using to run `mull-cxx`:

```
-test-framework=GoogleTest -mutators=math_add_mutator -compdb-path compile_commands.json -compilation-flags="-isystem /opt/llvm/7.0.0/include/c++/v1 -isystem /opt/llvm-7.0.0/lib/clang/7.0.0/include/ -isystem /opt/llvm-7.0.0/include/c++/v1 -isystem /usr/include" /opt/fmt/build.dir/bin/core-test
```

AlexDenisov:
GooD catch! We certainly need to cleanup the way it works and the getters for the original function and the original instruction. Current implementation is simply a bad design.

Here is what happens: we find a mutation point in a function X and store the to-be-mutated instruction as a class variable. Before mutation we create a copy of the function X, X_original, and replace the body of  X the following (pseudo-)code:
```
typedef void (*FunctionPointer)(int a, int b, int c);
FunctionPointer *X_pointer;
void X(int a, int b, int c) {
  X_pointer();
}
void X_original(int a, int b, int c) {
  // original implementation
}
void X_mutant1(int a, int b, int c) {
  // mutated implementation #1
}
void X_mutant2(int a, int b, int c) {
  // mutated implementation #2
}
```
This is done so that we can easily redirect the original call to a mutated function, i.e.:
```
for (Mutant &m : mutants) {
  X_pointer = m.PointerToMutatedFunction;
  X(a, b, c); // <- different functions called each time
}
```

When we replace the body of the function X, then the pointer to the original function is not correct (it should point to the X_original), and the pointer to the original value/instruction is not correct either: simply point to deallocated memory.

I think the solution for this problem is to update the pointer right after we do make the copy, or after we apply a mutation.

I hope the understanding of the problem is correct and it is a bit clearer what happens.

Issue #312, opened 2018-01-30
[WIP] Consistent Filters Proposal


stanislaw:
# Consistent Filters Proposal

Let's first accumulate all thoughts and use cases below:

## Thoughts

- Naming: we would love consistency and symmetry in config.yml and code.
- Some consistent solution for whitelisting and blacklisting for all of the options should be provided.
  - whitelisting on top of blacklisting i.e. whitelisting enables from `all - blacklisted` or from `all`?

### Focused run

- **whitelisting** 
  - I want to run only one or many specific tests or tests from one file (whitelisting)
- **blacklisting** 
  - I want to exclude one or many specific tests or tests from one file from consideration forever, and still be able to run specific tests in whitelisting mode.
  - I want to exclude certain functions from mutation search.

### Stray mutations

- We want to filter for `malloc`, `assert`, `printf` that would make any mutation operator skip these functions (Remove Void, Scalar Value, Negate Condition are the most popular).

AlexDenisov:
Also, filter for mutation points.

stanislaw:
@AlexDenisov I have added my own point:

> I want to exclude certain functions from mutation search.

Which filter do you mean? Please be specific.

AlexDenisov:
```yaml
exclude_mutations:
  - aes_cbc_40b24dcfdeb6ded614f9affafe78f3d4_0_0_10_replace_assignment_mutation_operator
```

AlexDenisov:
There was no discussion for the last year.

PR #551, opened 2019-08-16
Introduce Mutation Filters

AlexDenisov:
Wrap JunkDetector into the MutationFilter.


AlexDenisov:
Partially addresses #312

lib/Driver.cpp
```diff
@@ -1,3 +1,5 @@
+#include <utility>
```
AlexDenisov: CLion artifact, will remove in the upcoming PR.

lib/Driver.cpp
```diff
@@ -30,9 +32,7 @@ using namespace llvm::object;
 using namespace mull;
 using namespace std;
 
-Driver::~Driver() {
-  delete this->diagnostics;
-}
+Driver::~Driver() { delete this->diagnostics; }
```
AlexDenisov: I definitely want to revisit the clang-format one more time.

PR #556, opened 2019-08-19
Fix mutation elements over-reporting mutants

AlexDenisov:


lib/Reporters/MutationTestingElementsReporter.cpp
```diff
@@ -36,30 +36,17 @@ json11::Json createFiles(const Result &result,
   // on their source file.
   // Step 1: we create a map: "file => all of its mutation points".
   std::map<std::string, std::vector<MutationPoint *>> mutationPointsPerFile;
-  for (auto &mutationResult : result.getMutationResults()) {
-    MutationPoint *mutant = mutationResult->getMutationPoint();
-
+  for (auto mutant : result.getMutationPoints()) {
```
AlexDenisov: Here is the actual bug: a mutation point may have more than one mutation result. If we execute 5 different tests against one mutant, then we get 5 mutation results for this one mutation point.
Therefore, if we collect all mutation points for each mutation result, then we end up with 5 mutants, which are in fact 5 copies of the same mutant.

Correct behavior here is to go through mutation points since they are unique.

lib/Reporters/MutationTestingElementsReporter.cpp
```diff
@@ -36,30 +36,17 @@ json11::Json createFiles(const Result &result,
...
-          std::make_pair(sourceLocation.filePath,
-                         std::vector<MutationPoint *>());
-
-      mutationPointsPerFile.insert(pair);
-    }
-
-    std::vector<MutationPoint *> &currentFileMutationPoints =
-        mutationPointsPerFile.at(sourceLocation.filePath);
-    currentFileMutationPoints.push_back(mutant);
+    mutationPointsPerFile[sourceLocation.filePath].push_back(mutant);
```
AlexDenisov: There is no need to insert empty values into the `map`: `std` will instantiate an empty vector for you if it cannot find the corresponding entry.

lib/Reporters/MutationTestingElementsReporter.cpp
```diff
@@ -36,30 +36,17 @@ json11::Json createFiles(const Result &result,
...
   }
 
   // Step 2: Iterate through each file and dump the information about each
   // mutation points to its file's JSON entry.
-  for (const std::pair<std::string, std::vector<MutationPoint *>>
-           &fileMutationPoints : mutationPointsPerFile) {
+  for (auto &fileMutationPoints : mutationPointsPerFile) {
```
AlexDenisov: `clang-tidy` was complaining that there is an implicit conversion between `iterator` type and the `pair`. I just silenced the warning by using `auto`. Also, the explicit type is too verbose to my taste.

lib/Reporters/MutationTestingElementsReporter.cpp
```diff
@@ -76,9 +63,7 @@ json11::Json createFiles(const Result &result,
           sourceInfoProvider.getSourceInfo(mutationPoint);
 
       std::string status =
-          (killedMutants.find(mutationPoint) == killedMutants.end())
-              ? "Survived"
-              : "Killed";
+          (killedMutants.count(mutationPoint) == 0) ? "Survived" : "Killed";
```
AlexDenisov: The primary reason for the change: given that the intention here is `check if contains` I believe that `count` is more semantically correct than the `find`.
The secondary reason for the change: I didn't measure, but I assume that comparison of numbers (`== 0`) is more efficient than the comparison of iterators since the iterators are classes, which may involve either function calls, or heavy template code inlining.

lib/Reporters/MutationTestingElementsReporter.cpp
stanislaw: That was my point from the very beginning. It is good that now we are on the same page even though you point about the template-based code producing different results still applies.

I have reused this code from the IDEReporter without looking too much into how it should be done:

```
  for (auto &mutationResult : result.getMutationResults()) {
    auto mutant = mutationResult->getMutationPoint();
    auto &executionResult = mutationResult->getExecutionResult();
```

lib/Reporters/MutationTestingElementsReporter.cpp
stanislaw: 👍

nicojs:
Wow! Awesome to see this branch is merged! When will this be released? 😍

AlexDenisov:
I'm so excited about the elements, very nice visualization!
I plan to include it in the very next release, which should happen around the 1st of September.

@nicojs I'll definitely show you some results then 😸

PR #557, opened 2019-08-23
Junk Detection for the scalar value mutator

### LINKED ISSUE #484: Junk detection for scalar_value_mutator

AlexDenisov:
This is an initial implementation, it does not cover all the cases, but
should be a good start.

Closes #484.


AlexDenisov:
@stanislaw would appreciate if you can take a look.

stanislaw:
It looks great, Alex! 👍

PR #585, opened 2019-10-08
Split and parallelize mutation preparations

AlexDenisov:


AlexDenisov:
@stanislaw could you check the numbers, please?

PR #589, opened 2019-10-20
WIP: mull/AST: ASTFinder, ASTFilter, ASTMutations, ...

stanislaw:
This is going to be a long review round...

### DIFF
--- a/include/mull/AST/ASTTypes.h
+++ b/include/mull/AST/ASTTypes.h
@@ -15,10 +15,11 @@ enum TraverseMask: int {
 
   AND_OR = (1u << 0),
   BINARY_OP = (1u << 1),
-  SCALAR = (1u << 2),
-  VOID_CALL = (1u << 3),
+  NEGATE = (1u << 3),
+  SCALAR = (1u << 4),
+  VOID_CALL = (1u << 5),
 
-  ALL = AND_OR | BINARY_OP | SCALAR | VOID_CALL
+  ALL = AND_OR | BINARY_OP | NEGATE | SCALAR | VOID_CALL
 };
 struct MutationPointSourceInfo {
   int beginLine = -1;
--- a/include/mull/AST/ASTTypes.h
+++ b/include/mull/AST/ASTTypes.h
@@ -15,8 +15,9 @@ enum TraverseMask: int {
 
   BINARY_OP = (1u << 0),
   SCALAR = (1u << 1),
+  VOID_CALL = (1u << 2),
 
-  ALL = BINARY_OP | SCALAR
+  ALL = BINARY_OP | SCALAR | VOID_CALL
 };
 struct MutationPointSourceInfo {
   int beginLine = -1;
--- a/include/mull/AST/ASTTypes.h
+++ b/include/mull/AST/ASTTypes.h
@@ -1,0 +1,40 @@
+//
+// Created by Stanislav Pankevich on 2019-10-18.
+//
+
+#pragma once
+
+#include <clang/AST/Stmt.h>
+
+#include <string>
+#include <unordered_map>
+
+namespace mull {
+enum TraverseMask: int {
+  NONE = 0,
+
+  BINARY_OP = (1u << 0),
+  SCALAR = (1u << 1),
+
+  ALL = BINARY_OP | SCALAR
+};
+struct MutationPointSourceInfo {
+  int beginLine = -1;
+  int beginColumn = -1;
+  int endLine = -1;
+  int endColumn = -1;
+};
+
+using MutationLocationHash = int;
+
+using SourceFilePath = std::string;
+
+struct ASTMutation {
+  int line;
+  int column;
+  const clang::Stmt *const stmt;
+  ASTMutation(int line, int column, const clang::Stmt *const stmt)
+    : line(line), column(column), stmt(stmt) {}
+};
+using SingleASTUnitMutations = std::unordered_map<MutationLocationHash, ASTMutation>;
+} // namespace mull
--- a/include/mull/AST/ASTTypes.h
+++ b/include/mull/AST/ASTTypes.h
@@ -16,10 +16,11 @@ enum TraverseMask: int {
   AND_OR = (1u << 0),
   BINARY_OP = (1u << 1),
   NEGATE = (1u << 3),
-  SCALAR = (1u << 4),
-  VOID_CALL = (1u << 5),
+  REPLACE_CALL = (1u << 4),
+  SCALAR = (1u << 5),
+  VOID_CALL = (1u << 6),
 
-  ALL = AND_OR | BINARY_OP | NEGATE | SCALAR | VOID_CALL
+  ALL = AND_OR | BINARY_OP | NEGATE | REPLACE_CALL | SCALAR | VOID_CALL
 };
 struct MutationPointSourceInfo {
   int beginLine = -1;
--- a/include/mull/AST/ASTTypes.h
+++ b/include/mull/AST/ASTTypes.h
@@ -10,7 +10,7 @@
 #include <unordered_map>
 
 namespace mull {
-enum TraverseMask: int {
+enum TraverseMask : int {
   NONE = 0,
 
   AND_OR = (1u << 0),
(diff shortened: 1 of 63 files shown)


include/mull/MutationsFinder.h
```diff
@@ -20,7 +21,9 @@ class Testee;
 class MutationsFinder {
 public:
   explicit MutationsFinder(std::vector<std::unique_ptr<Mutator>> mutators,
-                           const Configuration &config);
+                           const Configuration &config,
+                           const FilePathFilter &filePathFilter,
+                           const InstructionFilter &instructionFilter);
```
stanislaw: Could we maybe merge these two filters into one and then do:

`shouldExcludePath` and `shouldExcludeInstruction`?

lib/AST/ASTFilter.cpp
```diff
@@ -0,0 +1,39 @@
...
+    return false;
+  }
+
+  int line = debugInfo.getLine();
+  int column = debugInfo.getCol();
+
+  const llvm::DILocation *diLocation = debugInfo.get();
+  std::string fileName = diLocation->getFilename();
+
+  if (filePathFilter.shouldSkip(fileName)) {
```
stanislaw: having it here does a great job of filtering!

lib/AST/ASTFinder.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+  TaskExecutor<mull::ASTSearchTask> finder(
+      "AST Search: looking for mutations in the source files",
+      program.bitcode(), mutationsAsVector, tasks);
+  finder.execute();
+
+  mull::ASTMutations mutations;
+
+  for (auto &mutationPair : mutationsAsVector) {
+    mutations.storage.emplace(mutationPair.first, mutationPair.second);
```
stanislaw: As discussed in person, the task executor only works with vectors so I do the mapping from and to `unordered_map`. Performance should not be affected seriously though.

Not sure if we really need to change the Task/Executor to support `unordered_map`.

lib/AST/ASTMutations.cpp
```diff
@@ -0,0 +1,72 @@
...
+    /// std::cout << "ASTMutations.count() one ast unit: " << pair.first << "\n";
+    count += storage.at(pair.first).size();
+  }
+  return count;
+};
+
+bool ASTMutations::locationExists(SourceFilePath sourceFile, int line,
+                                  int column) const {
+  assert(llvm::sys::fs::is_regular_file(sourceFile) ||
+         sourceFile == "input.cc");
```
stanislaw: "input.cc" is what you get by default when you create AST from a string using clang tooling.

I haven't found a better way to skip `input.cc` but at the same time assert that the paths are real when running real `mull-cxx`.

lib/AST/ASTMutations.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+  MutationLocationHash hash = locationHash(line, column);
+
+  assert(astUnitMutations.count(hash) > 0);
+
+  const ASTMutation &mutation = astUnitMutations.at(hash);
+  return mutation;
+}
+
+void ASTMutations::dump() const {
```
stanislaw: I believe that every unit of interest should have a `dump()` method because otherwise we keep repeating the same stuff for debugging all over and over again.

It can also be implemented as `dump(ostream&)` but I am not sure if we need to follow how LLVM does it.

lib/AST/ASTSearchTask.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    const llvm::Module *const module = bitcode->getModule();
+
+    if (pathFilter.shouldSkip(module->getSourceFileName())) {
+      continue;
+    }
+
+    /// How can I check that -g flag (debug info enable) was set, from llvm pass
+    /// https://stackoverflow.com/a/21713717/598057
+    if (module->getNamedMetadata("llvm.dbg.cu") == nullptr) {
```
stanislaw: I do believe that we need this kind of check, see #593. 

But maybe this could be moved even earlier, to be at the time when the bitcode has just finished loading.

lib/AST/ASTSearchTask.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    auto ast = astStorage.findASTByPath(sourceFileName);
+
+    ASTVisitor visitor(*ast, mutations, pathFilter, traverseMask);
+    visitor.TraverseDecl(ast->getASTContext().getTranslationUnitDecl());
+  }
+
+  //mutations.dump();
+
+  for (auto elem : mutations.storage) {
```
stanislaw: again, this is mapping `unordered_map -> vector -> unordered_map` because Executor only supports vectors.

lib/AST/ASTVisitor.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+  const std::string sourceLocation =
+    astUnit.getSourceManager().getFilename(exprLocation).str();
+
+  if (sourceLocation.empty()) {
+    /// TODO: asserts only?
+    return true;
+  }
+
+  if (filePathFilter.shouldSkip(sourceLocation)) {
```
stanislaw: This also provides a great benefit of filtering what is not in our files.

lib/AST/ASTVisitor.cpp
```diff
@@ -0,0 +1,220 @@
...
+    /// TODO: maybe there are more cases.
+    return;
+  }
+
+  if (!llvm::sys::fs::is_regular_file(sourceFilePath) &&
+      sourceFilePath != "input.cc") {
+    llvm::errs() << "error: ASTVisitor: invalid source file path: '"
+                 << sourceFilePath << "'\n";
+
+    exit(1);
```
stanislaw: In strict mode this is `exit()`, in the non-strict it is a warning and return

lib/Mutators/AndOrReplacementMutator.cpp
```diff
@@ -588,9 +588,9 @@ AND_OR_MutationType AndOrReplacementMutator::findPossibleMutationInBranch(
   return AND_OR_MutationType_None;
 }
 
-std::vector<MutationPoint *>
-AndOrReplacementMutator::getMutations(Bitcode *bitcode,
-                                      llvm::Function *function) {
+std::vector<MutationPoint *> AndOrReplacementMutator::getMutations(
+    Bitcode *bitcode, llvm::Function *function,
+    const mull::InstructionFilter &instructionFilter) {
```
stanislaw: All Mutators now get `instructionFilter` as a parameter. It is very important job that `instructionFilter` does, so I would be fine keeping it like this in case there is no better solution to pass it to Mutator in some other way.

lib/Mutators/CXX/TrivialCXXMutator.cpp
```diff
@@ -29,14 +29,19 @@ void TrivialCXXMutator::applyMutation(llvm::Function *function,
...
+std::vector<MutationPoint *> TrivialCXXMutator::getMutations(
+    Bitcode *bitcode, llvm::Function *function,
+    const mull::InstructionFilter &instructionFilter) {
   assert(bitcode);
   assert(function);
 
   std::vector<MutationPoint *> mutations;
 
   for (auto &instruction : instructions(function)) {
+    if (!instructionFilter.validMutation(instruction)) {
```
stanislaw: Great job of filtering using AST information.

lib/Reporters/ASTWhitelistSourceInfoProvider.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <iostream>
+
+using namespace mull;
+
+ASTWhitelistSourceInfoProvider::ASTWhitelistSourceInfoProvider(
+    ASTStorage &astStorage, ASTMutations &mutations)
+    : astStorage(astStorage), mutations(mutations) {}
+
+MutationPointSourceInfo
+ASTWhitelistSourceInfoProvider::getSourceInfo(MutationPoint *mutationPoint) {
```
stanislaw: This source info provider is implemented separately from the one used by Element because of the following:

- it uses `ASTMutations` to get the AST nodes and their locations collected by AST Search instead of using the information that is collected by Junk Detection.

lib/Reporters/IDEReporter.cpp
```diff
@@ -63,9 +63,9 @@ void IDEReporter::reportResults(const Result &result, const RawConfig &config,
 
   SourceManager sourceManager;
   for (auto mutant : result.getMutationPoints()) {
-    if (killedMutants.find(mutant) == killedMutants.end()) {
+    //if (killedMutants.find(mutant) == killedMutants.end()) {
```
stanislaw: Will be covered by #592 .

include/mull/AST/ASTTypes.h
```diff
@@ -0,0 +1,40 @@
...
+
+#pragma once
+
+#include <clang/AST/Stmt.h>
+
+#include <string>
+#include <unordered_map>
+
+namespace mull {
+enum TraverseMask: int {
```
stanislaw: I didn't find a better way of making it parameterize-able which mutations AST Search should look for. Without this functionality it is not possible to get just one mutation in certain test cases of the `ASTVisitor_*_Tests`.

stanislaw:
Closing this one if favour of #677 that is the first big step of splitting this one into very small pieces.

PR #595, opened 2019-10-22
Apply only mutations that are not junk

AlexDenisov:
This is how it works before the patch:

 - find all the mutations
 - put them into a bitcode file
 - apply each mutation added to the bitcode file

The problem is that whenever the mutation is considered as junk and
removed from the execution pipeline it is still applied as if it would
run.

The patch moves adding of mutants to the bitcode only after all the junk
is filtered out.

---

Numbers for 38k mutants only 19 of which are not filtered out:

Before: ~20 minutes

```
Cloning functions for mutation (threads: 4): 4/4. Finished in 1167456ms.
Removing original functions (threads: 4): 4/4. Finished in 32479ms.
```

After: ~100ms

```
Cloning functions for mutation (threads: 4): 4/4. Finished in 79ms.
Removing original functions (threads: 4): 4/4. Finished in 36ms.
```

### DIFF
--- a/include/mull/Bitcode.h
+++ b/include/mull/Bitcode.h
@@ -41,7 +41,6 @@ private:
   std::string uniqueIdentifier;
 
   std::map<llvm::Function *, std::vector<MutationPoint *>> mutationPoints;
-  std::mutex mutex;
 
   explicit Bitcode(std::unique_ptr<llvm::Module> module);
 };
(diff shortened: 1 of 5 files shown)


AlexDenisov:
All right, CI is red because it cannot install CMake. I'll re-try later.

PR #597, opened 2019-10-26
Rename Testee into ReachableFunction

AlexDenisov:
This is essentially a no-op, just better names for the `Testee` related code.


AlexDenisov:
Build on macOS is broken for whatever reason :(

PR #617, opened 2019-11-16
driver-cxx and FilePathFilter: whitelisting paths via -include-path

stanislaw:
[skip ci]

### DIFF
--- a/include/mull/Filters/FilePathFilter.h
+++ b/include/mull/Filters/FilePathFilter.h
@@ -3,7 +3,10 @@
 #include "mull/Filters/FunctionFilter.h"
 #include "mull/Filters/InstructionFilter.h"
 #include "mull/Filters/MutationFilter.h"
+
+#include <mutex>
 #include <regex>
+#include <unordered_map>
 #include <vector>
 
 namespace mull {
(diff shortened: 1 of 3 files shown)


include/mull/Filters/FilePathFilter.h
```diff
@@ -18,10 +20,16 @@ class FilePathFilter : public MutationFilter,
...
 
 private:
   bool shouldSkip(const mull::SourceLocation &location) const;
+  bool shouldSkip(const std::string &sourceFilePath) const;
+
+  std::vector<std::regex> includeFilters;
+  std::vector<std::regex> excludeFilters;
 
-  std::vector<std::regex> filters;
+  mutable std::unordered_map<std::string, bool> cache;
```
stanislaw: I am not sure about this `mutable` but otherwise we have to remove `const` from all of the declarations. I am fine with both options because this `FilePathFilter` class is very small is clear.

lib/Filters/FilePathFilter.cpp
```diff
@@ -23,18 +23,46 @@ bool FilePathFilter::shouldSkip(llvm::Instruction *instruction) const {
...
+        }
+      }
+    } else {
+      allow = false;
+
+      for (const auto &r : includeFilters) {
+        if (std::regex_search(sourceFilePath, r)) {
+          allow = true;
+        }
+        break;
```
AlexDenisov: I suppose it should be inside the `if`?

PR #619, opened 2019-11-16
mull-cxx: filter storage to allow more flexible creation of the filters

stanislaw:


tools/driver-cxx/driver-cxx.cpp
```diff
@@ -306,26 +308,36 @@ int main(int argc, char **argv) {
 
   auto sandbox = GetProcessSandbox(SandboxOption);
 
+  std::vector<std::shared_ptr<mull::Filter>> filterStorage;
```
stanislaw: I am under a strong impression that this storage can be encapsulated inside `mull::Filters`.

`mull::Filters` could then have a factory method for creating all filters from the CLI options and from needed dependencies such as JunkDetector.

Then the driver could be shorter and clear, with no details of filter creation.

tools/driver-cxx/driver-cxx.cpp
```diff
@@ -306,26 +308,36 @@ int main(int argc, char **argv) {
...
+    filePathFilter->exclude(regex);
   }
   for (const auto &regex : IncludePaths) {
-    filePathFilter.include(regex);
+    filePathFilter->include(regex);
   }
 
   if (junkDetectionEnabled) {
-    filters.mutationFilters.push_back(&junkFilter);
+    std::shared_ptr<mull::JunkMutationFilter> junkFilter(
```
stanislaw: This is a kind of change I will need for white search as well: junk filter can now be created inside a scope of `if (junkDetectionEnabled)` and it will survive the scope and will live as long as the filter storage.

Issue #629, opened 2019-11-27
YAML "unexpected scalar" Error when loading compilation database generated from Clang -MJ


Warwolt:
I'm running Mull 0.5.0 for Ubuntu 18.04 running on WSL1, and launched it with the -compdb-path option pointing to a compilation database json file generated by Clang. 

I'm compiling each c source-file using Clang 6.0.0-1ubuntu2, with the -MJ command for generating a json-file per object file (*.o.json). I then use Sed to stitch together all the individual json-files into a compilation database based on this [link](https://github.com/cquery-project/cquery/wiki/Compilation-database).

My current compilation-database looks like this (only one source-file currently under test):
```JSON
[
    {
        "directory": "/root/code/c/unit-test-experiments/mull_mutation_testing/build",
        "file": "../sw/src/millisec_timer_unix.c",
        "output": "/tmp/millisec_timer_unix-19be99.bc",
        "arguments": [
            "/usr/lib/llvm-6.0/bin/clang",
            "-xc",
            "../sw/src/millisec_timer_unix.c",
            "-c",
            "-Wall",
            "-Werror",
            "-fembed-bitcode=all",
            "-g",
            "-I",
            "../sw/inc",
            "-I",
            "../sw/test/fakes",
            "-I",
            "../lib/fakefunctionframework",
            "-o",
            "obj/millisec_timer_unix.o",
            "--target=x86_64-pc-linux-gnu"
        ]
    }
]
```

The command used to run Mull was:
```
mull-cxx -test-framework=GoogleTest -mutators=default -ld_search_path=/lib/x86_64-linux-gnu/ -ld_search_path=/usr/lib/x86_64-linux-gnu/ -ld_search_path=/usr/share/gdb/auto-load/usr/lib/x86_64-linux-gnu/ -exclude-path=lib -exclude-path=sw/test -exclude-path=googletest -compdb-path=./compile_commands.json ./test
```

The error given is:
```
mull-cxx -test-framework=GoogleTest -mutators=default -ld_search_path=/lib/x86_64-linux-gnu/ -ld_search_path=/usr/lib/x86_64-linux-gnu/ -ld_search_path=/usr/share/gdb/auto-load/usr/lib/x86_64-linux-gnu/ -exclude-path=lib -exclude-path=sw/test -exclude-path=googletest -compdb-path=./compile_commands.json ./test
Extracting bitcode from executable (threads: 1): 1/1. Finished in 23ms.
[...]

PR #631, opened 2019-11-27
LIT integration tests: boilerplate and the first test

stanislaw:


infrastructure/helpers/lit-tests.yaml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Run Mull examples
```
stanislaw: How can I run just this helper file to iterate on it locally?

tests-lit/tools/FileCheck.py
```diff
@@ -0,0 +1,385 @@
+#!/usr/bin/env python3
```
stanislaw: @AlexDenisov Are you comfortable with having FileCheck.py as source file, not as Python package for now?

tests-lit/tools/compilation_database.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/local/bin/python3
```
stanislaw: This is exactly what the #629 is about. I would be fine with having this for now. And then discarding it when Mull starts to support that compilation database format.

tests-lit/tools/FileCheck.py
stanislaw: Another option is to keep the binaries for now. Just let me know what you think.

AlexDenisov: Python source is definitely better than a binary.

infrastructure/helpers/lit-tests.yaml
AlexDenisov: Exactly as described in the scenario:
```
cd $SOURCE_DIR/tests-lit
export CLANG_EXEC=$LLVM_DIR/bin/clang++
make run_cxx
```

PR #634, opened 2019-12-01
Automate docs generation

AlexDenisov:


docs/generated/CLIOptions.rst
```diff
@@ -0,0 +1,168 @@
+--workers number		How many threads to use
+
+--dry-run		Skips real mutants execution. Disabled by default
```
stanislaw: indentation?

docs/conf.py
```diff
@@ -14,17 +14,25 @@
...
 
+import re
 import guzzle_sphinx_theme
 
 # -- Project information -----------------------------------------------------
 
-project = 'Mull'
-copyright = '2019, Alex Denisov'
-author = 'Alex Denisov'
+def project_version():
```
stanislaw: Great stuff

docs/generated/CLIOptions.rst
stanislaw: though you probably don't have control over this

AlexDenisov: It's automatically generated. We shouldn't care about the prettiness of the contents as long as sphinx can handle it correctly.

PR #658, opened 2020-02-10
CMake: tests-lit target for running LIT tests

stanislaw:


tests-lit/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+add_custom_target(tests-lit)
+
+set(CLANG_EXEC ${PATH_TO_LLVM}/bin/clang++)
+set(FILECHECK_EXEC filecheck)
```
stanislaw: What is the best way to check / ensure that `filecheck` exists and throw an error otherwise?

AlexDenisov: I think the 'right' way is to use [find_program](https://cmake.org/cmake/help/latest/command/find_program.html), but it's up to you.

tests-lit/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+add_custom_target(tests-lit)
```
AlexDenisov: What does it do?

stanislaw: It is a target that simply executes the custom command above. Do you see anything wrong with this approach?

AlexDenisov: Not necessarily wrong, but looks a bit suspicious to me, though it's just a gut feeling. I think it's fine.

tests-lit/CMakeLists.txt
stanislaw: As discussed in person, we leave it with `set`.

tests-lit/CMakeLists.txt
stanislaw: This has been changed.

PR #663, opened 2020-02-29
Junk detection: read compilation database flags from bitcode

stanislaw:


tests-lit/tests/junk_detection/04_junk_detection_using_bitcode_compilation_flags_via_record_command_line_flag/sample.cpp
```diff
@@ -0,0 +1,26 @@
...
+}
+
+/**
+; RUN: cd / && %CLANG_EXEC -fembed-bitcode -g -DFLAG=1 %s -o %s.exe
+; RUN: cd %CURRENT_DIR
+; RUN: (unset TERM; %MULL_EXEC -test-framework CustomTest -mutators=all -reporters=IDE -ide-reporter-show-killed %s.exe 2>&1; test $? = 0) | %FILECHECK_EXEC %s --strict-whitespace --match-full-lines --check-prefix=WITHOUT-RECORD-COMMAND-LINE
+; WITHOUT-RECORD-COMMAND-LINE-NOT:Found compilation flags in the input bitcode
+; WITHOUT-RECORD-COMMAND-LINE:{{^.*}}sample.cpp:5:13: warning: Survived: Remove Void Call: removed llvm.dbg.declare [remove_void_function_mutator]{{$}}
+
+; RUN: cd / && %CLANG_EXEC -fembed-bitcode -g -DFLAG=1 -grecord-command-line %s -o %s.exe
```
stanislaw: Looks like the `-grecord-command-line` flag is only supported starting from LLVM 8. We need to find a way to run this integration test on LLVM >= 8.

tools/driver-cxx/driver-cxx.cpp
```diff
@@ -136,6 +141,11 @@ int main(int argc, char **argv) {
   bool junkDetectionEnabled = false;
   std::string cxxCompilationFlags;
   std::string cxxCompilationDatabasePath;
+  if (!bitcodeCompilationFlags.empty()) {
+    diagnostics.info(std::string("Found compilation flags in the input bitcode\n"));
```
AlexDenisov: IIRC, we don't need extra `\n` here.

lib/BitcodeMetadataReader.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+std::map<std::string, std::string> mull::BitcodeMetadataReader::getCompilationDatabase(
+    std::vector<std::unique_ptr<mull::Bitcode>> &bitcode) {
+  std::map<std::string, std::string> bitcodeCompilationFlags;
+  if (bitcode.empty()) {
+    return bitcodeCompilationFlags;
+  }
+  for (auto &bitcodeModule: bitcode) {
+    llvm::Module *module = bitcodeModule->getModule();
+    std::string moduleFileName = module->getSourceFileName();
```
AlexDenisov: I'm not 100% sure that just taking the `sourceFileName` here is the right way to go. Maybe we should try picking the dir + path from debug info as well, and then fall back to the sourceFileName? I think this is how it's done in some other places.

tests-lit/tests/junk_detection/04_junk_detection_using_bitcode_compilation_flags_via_record_command_line_flag/sample.cpp
AlexDenisov: I'd suggest taking a look at `SourceLocation` class.

PR #666, opened 2020-03-07
Drop spdlog

AlexDenisov:


AlexDenisov:
The `flockfile` issue was not related to `spdlog` and was solved by switching back to the `pipe(2)` approach.

PR #669, opened 2020-03-08
Diagnostics: enable debug-level logging

stanislaw:


tools/driver-cxx/CLIOptions.cpp
```diff
@@ -136,6 +136,13 @@ opt<bool> tool::IDEReporterShowKilled(
   init(false),
   cat(MullCXXCategory));
 
+opt<bool> tool::DebugEnabled(
```
AlexDenisov: Please add this option to the `dumpCLIInterface` function below, and then run `build-docs` target.

This should update `CommandLineReference.rst` doc.

PR #677, opened 2020-03-11
"White" AST search: Add to Sub mutation

stanislaw:


lib/AST/ASTSearchTask.cpp
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    if (!llvm::sys::fs::is_regular_file(sourceFileName) && sourceFileName != "input.cc") {
+      diagnostics.error(std::string("ASTSearchTask: invalid source file path: ") + sourceFileName);
+    }
+
+    auto ast = astStorage.findAST(sourceFileName);
+
+    mutationUnits.push_back(std::make_pair(sourceFileName, SingleASTUnitMutations()));
+    SingleASTUnitMutations &singleUnitMutations = mutationUnits.back().second;
```
stanislaw: To me this looks a bit weird: I might be missing the proper C++ reference magic: when I want to both store the `SingleASTUnitMutations()` in the `mutationUnits` but also pass a reference to it to the `ASTVisitor`.

lib/AST/ASTVisitor.cpp
```diff
@@ -0,0 +1,94 @@
...
+  }
+  return clang::RecursiveASTVisitor<ASTVisitor>::VisitFunctionDecl(Decl);
+}
+
+bool ASTVisitor::VisitExpr(clang::Expr *expr) {
+  if (shouldSkipCurrentFunction) {
+    return true;
+  }
+
+  clang::SourceLocation exprLocation = expr->getBeginLoc();
```
stanislaw: LLVM 3.9 compains:

```
 /opt/llvm-3.9.0/bin/clang++  -DSPDLOG_COMPILED_LIB -I/home/runner/work/mull/mull/include -I/home/runner/work/mull/mull/LLVMCompatibility/3.9.x -I/home/runner/work/mull/mull/vendor/spdlog/include -isystem /opt/llvm-3.9.0/include -isystem /home/runner/work/mull/mull/vendor/LibEBC/lib/include -isystem /home/runner/work/mull/mull/vendor/libirm/include -isystem /home/runner/work/mull/mull/vendor -D_GLIBCXX_USE_CXX11_ABI=1 -O3    -std=c++11 -fPIC -fvisibility-inlines-hidden -fno-exceptions -Wunused -Werror -fno-rtti -MD -MT lib/CMakeFiles/mull.dir/AST/ASTVisitor.cpp.o -MF lib/CMakeFiles/mull.dir/AST/ASTVisitor.cpp.o.d -o lib/CMakeFiles/mull.dir/AST/ASTVisitor.cpp.o -c /home/runner/work/mull/mull/lib/AST/ASTVisitor.cpp
    /home/runner/work/mull/mull/lib/AST/ASTVisitor.cpp:31:46: error: no member named 'getBeginLoc' in 'clang::Expr'
      clang::SourceLocation exprLocation = expr->getBeginLoc();
                                           ~~~~  ^
```

include/mull/AST/ASTSearchTask.h
```diff
@@ -0,0 +1,33 @@
...
+
+  ASTSearchTask(mull::Diagnostics &diagnostics, mull::ASTStorage &astStorage,
+                mull::MutatorKindSet mutatorKindSet, mull::FilePathFilter &pathFilter);
+
+  void operator()(iterator begin, iterator end, Out &storage, mull::progress_counter &counter);
+
+private:
+  mull::Diagnostics &diagnostics;
+  mull::ASTStorage &astStorage;
+  mull::MutatorKindSet mutatorKindSet;
```
AlexDenisov: Just a reference should be enough here, right?

include/mull/AST/ASTSearchTask.h
```diff
@@ -0,0 +1,33 @@
...
+class progress_counter;
+
+class ASTSearchTask {
+public:
+  using In = const std::vector<std::unique_ptr<mull::Bitcode>>;
+  using Out = std::vector<std::pair<SourceFilePath, SingleASTUnitMutations>>;
+  using iterator = In::const_iterator;
+
+  ASTSearchTask(mull::Diagnostics &diagnostics, mull::ASTStorage &astStorage,
+                mull::MutatorKindSet mutatorKindSet, mull::FilePathFilter &pathFilter);
```
AlexDenisov: Do you need the explicit namespace here?

lib/AST/ASTFinder.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <cassert>
+#include <iostream>
+#include <regex>
+
+using namespace mull;
+
+ASTFinder::ASTFinder(mull::MutatorKindSet mutatorKindSet) : mutatorKindSet(mutatorKindSet) {}
+
+void mull::ASTFinder::findMutations(mull::Diagnostics &diagnostics, const mull::Configuration &config,
```
AlexDenisov: Same here: either drop `using namespace` or drop explicit namespaces here.

lib/AST/ASTMutationFilter.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+bool ASTMutationFilter::shouldSkip(MutationPoint *point) {
+  llvm::Instruction *instruction = llvm::dyn_cast<llvm::Instruction>(point->getOriginalValue());
+  assert(instruction);
+
+  const mull::MutatorKind mutatorKind = point->getMutator()->mutatorKind();
+
+  const llvm::DebugLoc &debugInfo = instruction->getDebugLoc();
+  if (!debugInfo) {
+    return false;
```
AlexDenisov: Shouldn't it be `true`?

lib/AST/ASTSearchTask.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+  std::vector<std::pair<SourceFilePath, SingleASTUnitMutations>> mutationUnits;
+
+  for (auto it = begin; it != end; it++, counter.increment()) {
+    const std::unique_ptr<mull::Bitcode> &bitcode = *it;
+
+    const llvm::Module *const module = bitcode->getModule();
+    std::string sourceFileName = module->getSourceFileName();
+    if (pathFilter.shouldSkip(sourceFileName)) {
+      diagnostics.info(std::string("ASTSearchTask: skipping path: ") + sourceFileName);
```
AlexDenisov: Are you sure you want to put it into the "main" communication channel?

lib/AST/ASTSearchTask.cpp
AlexDenisov: It looks legit :-D

lib/AST/ASTVisitor.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+void ASTVisitor::saveMutationPoint(mull::MutatorKind mutatorKind, const clang::Stmt *stmt,
+                                   clang::SourceLocation location) {
+
+  int beginLine = sourceManager.getExpansionLineNumber(location, nullptr);
+  int beginColumn = sourceManager.getExpansionColumnNumber(location);
+
+  std::string sourceFilePath = astUnit.getSourceManager().getFilename(location).str();
+  if (sourceFilePath.size() == 0) {
```
AlexDenisov: It is recommended to use `empty()` instead.

lib/AST/ASTVisitor.cpp
```diff
@@ -0,0 +1,99 @@
...
+    /// we reach here because of asserts()
+    /// TODO: maybe there are more cases.
+    return;
+  }
+
+  if (!llvm::sys::fs::is_regular_file(sourceFilePath) && sourceFilePath != "input.cc") {
+    diagnostics.error(std::string("ASTVisitor: invalid source file path: '") +
+                      sourceFilePath + "'\n");
```
AlexDenisov: The last `\n` is not necessary: the logger will add it.

tools/driver-cxx/CLIOptions.cpp
```diff
@@ -122,6 +122,11 @@ opt<SandboxKind> tool::SandboxOption(
     init(SandboxKind::Timer),
     cat(MullCXXCategory));
 
+opt<bool> tool::EnableAST(
+  "enable-ast", llvm::cl::Optional,
+  llvm::cl::desc("Enable \"white\" AST search (disabled by default)"),
+  llvm::cl::cat(MullCXXCategory), llvm::cl::init(false));
```
AlexDenisov: Also, include it into the list below and re-run `build-docs`.

tools/driver-cxx/driver-cxx.cpp
```diff
@@ -155,24 +159,29 @@ int main(int argc, char **argv) {
...
+    bitcodeCompilationFlagsAvailable = true;
   }
   if (!tool::CompilationDatabasePath.empty()) {
     cxxCompilationDatabasePath = tool::CompilationDatabasePath.getValue();
-    junkDetectionEnabled = true;
+    compilationDatabasePathAvailable = true;
   }
+  bool compilationDatabaseInfoAvailable = bitcodeCompilationDatabaseAvailable ||
+                                          compilationDatabasePathAvailable ||
+                                          bitcodeCompilationFlagsAvailable;
```
AlexDenisov: Could you explain the need for these three extra variables? It feels a bit too noisy and adds more mental overhead than needed.

tools/driver-cxx/driver-cxx.cpp
```diff
@@ -206,7 +215,23 @@ int main(int argc, char **argv) {
...
+  if (tool::EnableAST && compilationDatabaseInfoAvailable) {
+    std::vector<mull::MutatorKind> mutationKinds;
+    for (auto &mutator : mutatorsOptions.mutators()) {
+      mutationKinds.push_back(mutator->mutatorKind());
+    }
+
+    mull::MutatorKindSet mutatorKindSet = mull::MutatorKindSet::create(mutationKinds);
+
+    mull::ASTFinder astFinder(mutatorKindSet);
+    astFinder.findMutations(diagnostics, configuration, program, *filePathFilter, astStorage);
```
AlexDenisov: Eventually, you may want to move it to a later stage and run only against functions covered with tests.

lib/AST/ASTVisitor.cpp
AlexDenisov: Btw, if it's only LLVM 3.9 why `LLVM_VERSION_MAJOR >= 7`?

lib/AST/ASTSearchTask.cpp
stanislaw: Thanks for confirming!

lib/AST/ASTVisitor.cpp
stanislaw: I think I simply got confused by the feedback from GitHub Actions and also misinformed you: the LLVM 3.9 simply gets run first so it was the first to be seen failed, but the actual edge is between 6 and 7.

include/mull/AST/ASTSearchTask.h
stanislaw: If you look at how this mutator kind set is created in driver-cxx.cpp you will see that it is a set of mutator kinds that is created on stack with `if (tool::EnableAST` branch. Within the current implementation we could make it a reference but I think it should either be like it is now or a && move.

I expected a review comment where you would suggest me to move the AST search part to the driver so there it would make even more sense. If we don't move it to the driver I can make it a reference. Please advise.

include/mull/AST/ASTSearchTask.h
stanislaw: Changed.

include/mull/AST/ASTSearchTask.h
stanislaw: Actually I have changed it to references for now. It is done.

lib/AST/ASTFinder.cpp
stanislaw: Done.

lib/AST/ASTMutationFilter.cpp
stanislaw: Good catch. It was a move from the old branch.

lib/AST/ASTSearchTask.cpp
stanislaw: Of course not! Thanks!

tools/driver-cxx/driver-cxx.cpp
stanislaw: I fully agree. Two additional comments:

1) I expected that you would suggest me to move this to the Driver code anyway and I am happy to do that later.
2) Given that we will use white search combined with whitelisting of the paths the difference between what is being run and where the mutations are found should be as dramatic as before.

tools/driver-cxx/driver-cxx.cpp
stanislaw: This is the only comment where I want to make an argument for actually doing this.

We can make the variables a little bit shorter but they actually make making of decisions below very clear:

We only enter the "Junk detection" decision only if `compilationDatabaseInfoAvailable` and we enter the "White search" decision if `EnableASTSearch && compilationDatabaseInfoAvailable`.

Otherwise I would have to introduce a variable called `whiteSearchIsEnabled` and put it where `junkDetectionEnabled` is set to `true` and we would have a duplication of both variables set to true to make their decisions.

This makes sense to me. For you?

AlexDenisov: It does make sense to me, though there is a small misunderstanding on either side.
What I meant is to basically rename `junkDetectionEnabled` into `compilationDatabaseInfoAvailable` and drop the unnecessary variables. Then the same logic would be preserved, right?

I'm not going to die on this hill, so if having three extra variables works better for you, then I don't have any objections :)

PR #682, opened 2020-03-19
ASTVisitor: recognize possible Sub to Add mutation

stanislaw:


lib/AST/ASTVisitor.cpp
```diff
@@ -49,12 +54,16 @@ bool ASTVisitor::VisitExpr(clang::Expr *expr) {
...
-    if (binaryOperator->getOpcode() == clang::BO_Add &&
-        mutatorKindSet.includesMutator(mull::MutatorKind::CXX_AddToSub)) {
-      clang::SourceLocation binaryOperatorLocation = binaryOperator->getOperatorLoc();
-      saveMutationPoint(mull::MutatorKind::CXX_AddToSub, binaryOperator, binaryOperatorLocation);
-      return true;
+    clang::SourceLocation binaryOperatorLocation = binaryOperator->getOperatorLoc();
+    for (const std::pair<clang::BinaryOperator::Opcode, mull::MutatorKind> &mutation :
+         BINARY_MUTATIONS) {
+      if (binaryOperator->getOpcode() == mutation.first &&
+          mutatorKindSet.includesMutator(mutation.second)) {
```
AlexDenisov: Just a side note: it would probably be a good idea to not even enter this method if the `mutatorKindSet` does not include available mutators.

PR #690, opened 2020-03-24
Decrease package/binary size

AlexDenisov:
This PR drops the size of the package (deb/zip) and the binary (mull-cxx).
Here are the numbers from macOS/release build:
Package size: 20M vs ~16M~ 18M.
~Binary size: 57M vs 45M.~

Ok, we cannot strip symbols because JIT fails to find the symbols then.
But the package is still smaller because we drop libirm and libmull.a.


AlexDenisov:
Ubuntu package shrunk by ~2.5Mb as well.

PR #698, opened 2020-03-28
ASTVisitor: recognize "!a -> !!a"

stanislaw:


include/mull/Mutators/NegateConditionMutator.h
```diff
@@ -45,5 +45,6 @@ class NegateConditionMutator : public Mutator {
 
 private:
   std::vector<std::unique_ptr<irm::IRMutation>> lowLevelMutators;
+  irm::NegateXORReplacement negateXorMutator;
```
stanislaw: This looks a bit unusual since NegateXOR is rather different compared to the `lowLevelMutators`.

Does it mean that we have to introduce another high-level Mutator?

tests-lit/tests/white-ast-search/experimental/negate/02_unary_to_unary_negated/sample.cpp
```diff
@@ -0,0 +1,25 @@
...
+; RUN: cd %CURRENT_DIR
+; RUN: sed -e "s:%PWD:%S:g" %S/compile_commands.json.template > %S/compile_commands.json
+; RUN: (unset TERM; %MULL_EXEC -debug -enable-ast -test-framework CustomTest -mutators=negate_mutator -reporters=IDE -ide-reporter-show-killed -compdb-path %S/compile_commands.json %s.exe 2>&1; test $? = 0) | %FILECHECK_EXEC %s --strict-whitespace --match-full-lines
+; CHECK-NOT:{{^.*[Ee]rror.*$}}
+; CHECK-NOT:{{^.*[Ww]arning.*$}}
+
+; CHECK:[info] AST Search: looking for mutations in the source files (threads: 1)
+; CHECK:[debug] AST Search: recording mutation "Negate": {{.*}}sample.cpp:2:10
+
+; TODO: We don't support this mutation on the LLVM IR level where there is a
```
stanislaw: This is still a TODO. We need to discuss which low-level mutation this should be:

```
; Function Attrs: nounwind ssp uwtable
define zeroext i1 @_Z6isTrueb(i1 zeroext) #0 !dbg !7 {
  %2 = alloca i8, align 1
  %3 = zext i1 %0 to i8
  store i8 %3, i8* %2, align 1
  call void @llvm.dbg.declare(metadata i8* %2, metadata !11, metadata !12), !dbg !13
  %4 = load i8, i8* %2, align 1, !dbg !14
  %5 = trunc i8 %4 to i1, !dbg !14
  ret i1 %5, !dbg !15
}
...
!14 = !DILocation(line: 2, column: 10, scope: !7)
```

line: 2, column: 10 point here:

```c
  return a;
         ^
```

PR #699, opened 2020-03-28
Update docs

AlexDenisov:


AlexDenisov:
I canceled builds since there is nothing to validate in this PR.

PR #700, opened 2020-03-29
Enable LTO for release builds

AlexDenisov:
It shrinks the binary size even further


AlexDenisov:
Ok, it would require more work on Linux, so I'll postpone until some later point.

PR #701, opened 2020-04-02
 MutationPoint: dump() method to simplify debugging 

stanislaw:


lib/MutationPoint.cpp
```diff
@@ -127,33 +137,47 @@ std::string MutationPoint::getMutatorIdentifier() const {
...
-  return originalFunction->getName().str() + "_" +
-         bitcode->getUniqueIdentifier() + "_original";
+  return originalFunction->getName().str() + "_" + bitcode->getUniqueIdentifier() + "_original";
+}
+
+std::string MutationPoint::dump() const {
+  std::stringstream ss;
+  ss << "Mutation Point: " << getMutator()->getUniqueIdentifier() << " "
+     << getSourceLocation().filePath << ":" << getSourceLocation().line << ":"
+     << getSourceLocation().column;
```
AlexDenisov, line 181: Could've been `getSourceLocation().dump()`, right? ;)

stanislaw: Good idea. I am all in for having focused `.dump()` methods.

PR #702, opened 2020-04-05
MutationPoint: dump mutation's source file context

stanislaw:


include/mull/Reporters/SourceManager.h
```diff
@@ -18,6 +18,7 @@ struct LineOffset {
 class SourceManager {
 public:
   std::string getLine(const SourceLocation &location);
+  size_t getLinesNumber(const SourceLocation &location);
```
AlexDenisov, line 21: Why plural?

lib/Reporters/SourceManager.cpp
```diff
@@ -12,7 +12,7 @@ SourceManager::~SourceManager() {
 
 std::string SourceManager::getLine(const SourceLocation &location) {
   LineOffset &lineOffset = getLineOffset(location);
-  assert(location.line < lineOffset.offsets.size());
+  assert(location.line <= lineOffset.offsets.size());
```
AlexDenisov, line 15: Could you elaborate on this? Accessing `size()` element would be out of bounds, right?

include/mull/Reporters/SourceManager.h
stanislaw: Changed to `getNumberOfLines`.

lib/Reporters/SourceManager.cpp
stanislaw: As discussed in person, the issue was related to the fact that the code didn't support getting the information for last lines in files.

PR #706, opened 2020-04-09
MutationTestBed class: encapsulate creation of mutation content for unit tests

stanislaw:
I am not sure about the names of the classes but it works perfectly :)

### DIFF
--- a/tests/Helpers/MutationArtefact.cpp
+++ b/tests/Helpers/MutationArtefact.cpp
@@ -1,0 +1,36 @@
+#include "MutationArtefact.h"
+
+#include "mull/Bitcode.h"
+
+using namespace mull;
+using namespace mull_test;
+
+MutationArtefact::MutationArtefact(mull::SingleASTUnitMutations astMutations,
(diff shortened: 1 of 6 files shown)


tests/Helpers/MutationArtefact.cpp
```diff
@@ -0,0 +1,36 @@
...
+    junkMutationPoints(junkMutationPoints), bitcode(std::move(bitcode)),
+    context(std::move(context)) {}
+
+MutationArtefact::~MutationArtefact() {
+  /// We need to enforce the order of deallocation:
+  /// First we deallocate the LLVM bitcode and then the underlying LLVM context.
+  /// Doing this explicitly in order to avoid implicit deallocation of
+  /// MutationArtefact's members.
+  bitcode.reset();
+  context.reset();
```
AlexDenisov, line 23: I think you can just rearrange the class fields.

stanislaw: True. I will preserve a comment and simply put it above the members.

PR #711, opened 2020-04-17
ASTVisitor: recognize scalars in return statements

stanislaw:
Depends on #712 and #713.

### DIFF
--- a/lib/AST/ASTScalarMutationMatcher.cpp
+++ b/lib/AST/ASTScalarMutationMatcher.cpp
@@ -1,0 +1,51 @@
+#include "mull/AST/ASTScalarMutationMatcher.h"
+
+#include "mull/AST/MullClangCompatibility.h"
+#include "mull/Program/Program.h"
+
+#include <clang/AST/RecursiveASTVisitor.h>
+#include <clang/Basic/SourceLocation.h>
+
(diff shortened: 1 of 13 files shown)


lib/AST/ASTScalarMutationMatcher.cpp
```diff
@@ -0,0 +1,61 @@
...
+  return false;
+}
+
+ASTScalarMutationMatcher::ASTScalarMutationMatcher(clang::ASTContext &context)
+    : context(context) {}
+
+bool ASTScalarMutationMatcher::isMutableExpr(const clang::Expr &expr,
+                                             const clang::Stmt **mutableParentStmt,
+                                             clang::SourceLocation *mutationLocation) {
+  std::cout << "hello\n";
```
AlexDenisov, line 31: 👋

stanislaw: The true story behind this :)

I was creating this class from ASTVisitor that had `diagnostics` as class member. At some point I got compiler warnings that that field was not used. I decided to add this hello to just make compiler happy. Then I realized that I no longer need diagnostics there but I still decided to keep this placeholder for debugging. And then I forgot about it. It is not that simple :)

Issue #528, opened 2019-06-12
Could not find dynamic library


1BE6-ankit:
I have a docker setup for compiling my project. I am able to compile and run the project, but upon invoking mull I get the following error:
```
Extracting bitcode from executable (threads: 1): 1/1. Finished in 1ms.
Loading bitcode files (threads: 2): 2/2. Finished in 11ms.
Could not find dynamic library: libpthread.so.0
Could not find dynamic library: libstdc++.so.6
Could not find dynamic library: libm.so.6
Could not find dynamic library: libgcc_s.so.1
Could not find dynamic library: libc.so.6
Compiling instrumented code (threads: 2): 2/2. Finished in 10ms.
Loading dynamic libraries (threads: 1): 1/1. Finished in 0ms.
Searching tests (threads: 1): 1/1. Finished in 0ms.
Preparing original test run (threads: 1): 1/1. Finished in 3ms.
Running original tests (threads: 2): 0/2FactorialTest.NaturalCase failed: Crashed
FactorialTest.BaseCase failed: Crashed
Running original tests (threads: 2): 2/2. Finished in 191ms.
No mutants found. Mutation score: infinitely high

Total execution time: 218ms
```

My docker image looks like this:
```
FROM ubuntu:18.04

RUN mkdir -p /tmp/home
ENV HOME /tmp/home

RUN apt-get update && apt-get upgrade -y \
    && apt-get install -y git build-essential software-properties-common\
    && apt-get install -y vim wget

RUN apt-get update

RUN apt-get install -y python3-pip \
    && pip3 install conan

RUN apt-get update

## cpp unit
# RUN apt-get install -y autotools-dev automake libtool
RUN apt-get install -y libcppunit-dev  

# for mull
RUN apt-get install -y sqlite3 libsqlite3-dev libncurses5-dev libxml2-dev uuid-dev libstdc++-6-dev
```

To setup LLVM and mull I do the following:
```
export PATH="$PATH:/usr/src/llvm/bin"
cat > llvm.conf << "HEREDOC"
/usr/src/llvm/lib
HEREDOC

mv llvm.conf /etc/ld.so.conf.d/
ldconfig

export CC=/usr/src/llvm/bin/clang  
export CXX=/usr/src/llvm/bin/clang++

apt-get install -y cmake
apt-get update

dpkg -i setup/libicu55_55.1-7_amd64.deb
dpkg -i setup/Mull-0.3.0-LLVM-8.0-ubuntu-16.04.deb
```

[...]

PR #716, opened 2020-04-19
Enable LIT tests on CI for Linux

### LINKED ISSUE #715: Enable LIT tests on CI for Linux
Currently, LIT tests are only run on macOS.

stanislaw:
- [x] pip3 is missing in Ubuntu Linux playbook
- [x] `[warning] Could not find dynamic library: libstdc++.so.6`
- [x] `error: no member named 'abort' in namespace 'std'`

And after the libc fix:

- [x] `[warning] Could not find dynamic library: libc++.so.1`

Closes #715

### DIFF
--- a/infrastructure/ubuntu-playbook.yaml
+++ b/infrastructure/ubuntu-playbook.yaml
@@ -25,6 +25,7 @@
       - uuid-dev
       - binutils
       - liblzma-dev
+      - python3-pip
 
   vars_files:
     - helpers/variables.yaml
@@ -58,6 +59,8 @@
     - name: Build Mull
       include: helpers/build-mull.yaml
 
+    - name: LIT Tests
+      include: helpers/lit-tests.yaml
+
     - name: Integration Tests
       include: helpers/integration-tests.yaml
-
--- a/tests-lit/CMakeLists.txt
+++ b/tests-lit/CMakeLists.txt
@@ -14,8 +14,9 @@ else()
 endif()
 
 set(TEST_CXX_FLAGS
-  "-isystem ${PATH_TO_LLVM}/include/c++/v1 -isystem ${PATH_TO_LLVM}/lib/clang/${LLVM_VERSION}/include"
+  "-stdlib=libc++ -isystem ${PATH_TO_LLVM}/include/c++/v1 -isystem ${PATH_TO_LLVM}/lib/clang/${LLVM_VERSION}/include"
 )
+
 if (SYSROOT)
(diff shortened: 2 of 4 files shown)


stanislaw:
The current issue on Linux is exactly the one that was reported in #528 .

Adding the following Mull flags fixes this for me in Docker built with our provisioning files:

```
mull ... -ld-search-path=/usr/lib/x86_64-linux-gnu -ld-search-path=/lib/x86_64-linux-gnu ...
```

based on the following paths:

```
root@7f0e40e2d86e:/opt/build.mull.release.dir# ldd ../mull/tests-lit/tests/white-ast-search/scalar/01_return_value/sample.cpp.exe
	linux-vdso.so.1 =>  (0x00007fff3f9fc000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fc1f7325000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc1f701c000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fc1f6e06000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc1f6a3c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc1f76a7000)
```

--- 

The easiest solution I see would be to switch to using the %mull variable where we would mix in the portable flags or introduce another %mull-portable-flags variable where we would put this `-ld-search-path` stuff.

@AlexDenisov let me know what you think.

AlexDenisov:
The easiest solution would be to add those flags as is. Currently, they only effective on Linux, so adding other configuration options might be an overkill for this situation.

stanislaw:
> The easiest solution would be to add those flags as is.

What do you mean exactly?

Keep in mind that we have to do this in each and every lit test, this is why I am thinking of abstracting them with a variable like `%mull-portable-flags` in `lit.cfg`.

AlexDenisov:
> What do you mean exactly?

```
mull ... -ld-search-path=/usr/lib/x86_64-linux-gnu -ld-search-path=/lib/x86_64-linux-gnu ...
```

> Keep in mind that we have to do this in each and every lit test, this is why I am thinking of abstracting them with a variable like %mull-portable-flags in lit.cfg.

I think it should rather belong to the test, and not to the 'global' setup. But I don't mind if you put it into the config file, you decide.

stanislaw:
OMG!

PR #720, opened 2020-04-24
ASTVisitor: recognize scalars in function call arguments

stanislaw:


tests/Mutations-E2E/Scalar/04_Mutation_Scalar_CallArgument_Test.cpp
```diff
@@ -0,0 +1,61 @@
...
+using namespace llvm;
+
+static FilePathFilter nullPathFilter;
+
+static const std::string testCode = std::string(R"(///
+int dummy(int a) {
+  return a;
+}
+int caller() {
+  return dummy(0);
```
AlexDenisov, line 26: What is the expected behavior in case there are two arguments, e.g. `dummy(0, 5)`? I'm afraid this is not supported by Mull in the current implementation.

stanislaw:
Currently Mull only supports having instruction address in the mutation information so it does not allow us to store additional information, such as the number of a function argument as this type of mutation requires.

We have considered the idea of adding a generic `payload` field to the mutation information that would accommodate for having the `number of argument` information and we also agreed that we are not going to implement it now. 

Closing this until there is more need for having a more detailed information attached to mutation points.

PR #730, opened 2020-06-22
Fix incorrect extra qualification

MaskRay:


AlexDenisov:
Thanks for the patch @MaskRay! The CI failure seems to be unrelated to your changes, but I'll fix it while fixing the #731 and then merge your PR.

PR #735, opened 2020-06-26
SQLite tutorial

AlexDenisov:


docs/SQLiteReport.rst
```diff
@@ -0,0 +1,261 @@
+Working with SQLite report
+==========================
+
+From the very beginning, we didn't want to impose our vision on treating the results of mutation testing. Some people do not care about the mutation score, while others do care, but want to calculate it slightly differently.
+
+To solve this problem, Mull splits execution and reporting into separate phases.
+What Mull does is apply mutation testing on a program, collect as much information as possible, and then pass this information to one of several reporters.
```
stanislaw, line 7: Maybe `does is as follows: ` otherwise, it reads a bit short.

docs/SQLiteReport.rst
```diff
@@ -0,0 +1,261 @@
+Working with SQLite report
+==========================
+
+From the very beginning, we didn't want to impose our vision on treating the results of mutation testing. Some people do not care about the mutation score, while others do care, but want to calculate it slightly differently.
```
stanislaw, line 4: Looking through the whole diff: a width limit is used sometimes and sometimes is not. My preference for RST would still remain 80, but feel free to use your best judgment.

AlexDenisov: I put the doc into Grammarly before publishing and it forced me to remove all the new-lines, I should've put them back though.

I'll do another round of cleanups in a separate PR and I'll go over all the other documents as well.

PR #742, opened 2020-08-27
cmake: Cache PATH_TO_LLVM variable

gebart:
Avoids the need to specify it again if we run `cmake .` from the build directory.


AlexDenisov:
Nice one, thanks @gebart.
There is something odd with the CI, I'll look into it and merge the PR right after that.

AlexDenisov:
Actually, I think that the failures are good - if the PR is coming from the outside of Mull repo, then we should not publish a package.
Otherwise, nothing will stop a potential malefactor from inserting malicious code and opening PR. If that happens, then we publish the malicious package under Mull repo on Bintray, potentially exposing the users to a threat.

I'll merge this as-is and then add this exception to the CI action.

Issue #743, opened 2020-08-27
Crash after `JIT engine could not resolve the following symbols: ___emutls_get_address`


gebart:
This crash happens when trying to start any test program that contains at least one thread local variable. 
The cause of this is that when recompiling the bitcode, all TLS accesses are converted to using TLS emulation. I don't have much knowledge of this area but from what I have managed to find online, this behavior of the JIT engine should be similar to using the `-femulated-tls` flag with clang. Forcing TLS emulation off is possible by modifying the calls to the EngineBuilder in InstrumentedCompilationTask.cpp, but that seems to make the recompilation crash instead.

### Minimal example test subject which shows the problem

```c++
int main(int argc, char **argv) {
  thread_local static int c = 0;
  return c;
}
```

### Example run
```
❯ mull-cxx --debug --test-framework=CustomTest --sandbox=None ./build/HelloWorldTests
[debug] Diagnostics: Debug Mode enabled. Debug-level messages will be printed.
[info] Extracting bitcode from executable (threads: 1)
       [################################] 1/1. Finished in 2ms
[info] Loading bitcode files (threads: 1)
       [################################] 1/1. Finished in 12ms
[info] Compiling instrumented code (threads: 1)
       [################################] 1/1. Finished in 11ms
[info] Loading dynamic libraries (threads: 1)
       [################################] 1/1. Finished in 1ms
[info] Searching tests (threads: 1)
       [################################] 1/1. Finished in 0ms
[info] Preparing original test run (threads: 1)
[warning] JIT engine could not resolve the following symbols:
___emutls_get_address

       [################################] 1/1. Finished in 0ms
[info] Running original tests (threads: 1)
[1]    43292 segmentation fault  mull-cxx --debug --test-framework=CustomTest --sandbox=None
```

Running with sandbox enabled just shows that the test subject has crashed:
```
[warning] Original test failed
test: main
status: Crashed
stdout: ''
stderr: ''
```

### Platform

OS: MacOS 10.15.6 build 19G73
LLVM: clang+llvm-9.0.0-x86_64-darwin-apple from [llvm.org](https://releases.llvm.org/download.html#9.0.0)  (git://github.com/llvm/llvm-project.git 0399d5a9682b3cef71c653373e38890c63c4c365), Target: x86_64-apple-darwin19.6.0

### Analysis

[...]

PR #744, opened 2020-08-28
Add debug logging for object loading

gebart:
This adds debug level messages for loading libraries and objects, and for resolving external references between objects and libraries.
Useful to better see what is going on inside mull-cxx during the prepare phase. This was written while debugging #743.


AlexDenisov (approved):

gebart:
This does increase console spam a lot when running mutations with --debug since the lookups need to be done for each mutation run.

AlexDenisov:
Ok, there is another CI related issue that I somehow missed 😞

PR #745, opened 2020-09-01
emutls: Add helper shared library for TLS test subjects

gebart:
Fix for #743 

Add LLVM sources for `__emutls_get_address` and build `libmull-tls.dylib` which can be used to avoid missing symbol errors.

### To do:

- [ ] Add documentation on how to invoke `mull-cxx -ld-preload` for subjects using `thread_local` variables


AlexDenisov:
I'm wondering if it's a good idea to just supply something like `libmull-tls.so/dylib` (or even `libmull-rt.so/dylib`) and add an option to the driver (something like `-dylibs`) and ask the user to specify this option manually?
Then it would be something like:
```
mull-cxx ... -dylibs=/usr/lib/libmull-tls.so
```
and the `-dylibs` will just add whatever is passed to the `dynamicLbraries`.

Automatic resolution (as implemented in this draft) is nice, but the users would have to specify the library search paths. On one hand, if they are providing these options already then it's very convenient since no changes needed. On the other hand, if they have to specify the search paths, then it's just easier to specify the full path to the file.

We can probably think of several options, e.g.:
1. Add an extra flag for dynamic libraries.
2. Automatic lookup in the library search paths (maybe hidden behind `-enable-tls` flag).
3. Implement both 1 and 2? :)

Does anything of this make sense to you @gebart? :)

gebart:
@AlexDenisov thanks for the feedback. I agree that using a command line switch is cleaner.

I like your proposals, how about this implementation? (1+2 above)

- Add a command line arg, `-ld-preload=foo.so` (for similarity with `LD_PRELOAD` env var)
- If a file with the given path is found, load that library, otherwise use `-ld-search-path` to resolve the library file path.

The name libmull-tls is a much better description than libemutls, I will update this PR with that.

AlexDenisov:
@gebart I cannot stress enough how much I appreciate your work on this problem. In the end, we decided to not use JIT anymore, so this fix is no longer needed. I'm closing this PR hoping that the new release solves this issue and you have a more smooth integration into your project.

I'm looking forward for your further feedback.

PR #746, opened 2020-09-01
TestFrameworks: Add Boost.Test support for discovering test cases

gebart:
Add heuristics for guessing available test cases for Boost.Test binaries. This uses the symbol naming and namespaces to make a best guess of the available test cases, which works fine as long as the developer is using the boost provided macros for defining test cases and suites. Doing a real detection that covers all corner cases around manually registered tests would pretty much require running the test binary and extracting the list from internals after the program has started.

To do:
 - [x] add unit test
 - [x] remove some debug spam
 - [x] Handle templated test cases (might be affected by `-fno-rtti`)


AlexDenisov:
> Doing a real detection that covers all corner cases around manually registered tests would pretty much require running the test binary and extracting the list from internals after the program has started.

There is exactly the same problem with GoogleTest: if it's not a simple or more-or-less simple test, then we are screwed. So far the strategy was to ignore such edge cases, but it may change :)

gebart:
I think this is ready for review now.

AlexDenisov (approved):
I really like the idea of faking the test framework here. I think this is a great way to adopt more test frameworks in the future, especially if we want to support many versions of the same test framework.
Thank you, @gebart!

gebart:
TBH the effort to fake the test framework was a lot bigger than I originally expected. I believe that other frameworks may be lower effort though because, Boost...

PR #747, opened 2020-09-02
FilePathFilter: Allow combined include-path + exclude-path

gebart:
Include will be applied first, then excludes.

If only include-path or only exclude-path is given then the behavior is unchanged from before.

This is useful if test code lives in the same tree as the code under test, so you can do `mull-cxx --include-path=mymodule --exclude-path=tests`


AlexDenisov (approved):

AlexDenisov:
Thanks, @gebart. If you don't mind I'd like to take over this PR and push couple more commits on top to debug and fix the CI failure.

gebart:
Sure, feel free to update my branch with your changes.

gebart:
rebased to retrigger CI

PR #751, opened 2020-09-16
cmake: Link against native codegen and asmparser libs.

meklort:
This fixes a linking issue when building under ppc64le.


AlexDenisov:
The CI failure is unrelated to the change, so I'm merging this one as is. Thanks for the patch @meklort!

PR #756, opened 2020-09-22
Add command line option -ld-preload

gebart:
As suggested in https://github.com/mull-project/mull/pull/745#issuecomment-694489990
This adds a new CLI option `-ld-preload` for providing preloaded dynamic libraries. The primary use case is to override libraries for test binaries which fail when recompiled, in particular those using C++ `thread_local` variables.

Forked off from #745 to make reviewing easier.


AlexDenisov (approved):

AlexDenisov:
Travis's failure is fixed here #753, merging this one despite the red build.

Issue #657, opened 2020-02-10
undefined reference to `typeinfo for irm::CmpInstPredicateReplacement'


chgans:
Hi there,

I'm trying to build 0.6.1 tag from git in an Ubuntu-18.04/amd64 docker container.

I've installed LLVM using their convenience apt repo (/usr/lib/llvm-9).

```
RUN apt-get install --yes sqlite3 libsqlite3-dev git libxml2-dev uuid-dev liblzma-dev binutils ncurses-dev libz-dev \
    && cd /usr/src \
    && git clone https://github.com/mull-project/mull.git \
    && cd mull \
    && git checkout 0.6.1 \
    && git submodule init \
    && git submodule update \
    && mkdir build \
    && cd build \
    && export LLVM_ROOT=/usr/lib/llvm-9/ \
    && export PATH=$LLVM_ROOT/bin/:$PATH \
    && export CC=clang-9 \
    && export CXX=clang++-9 \
    && cmake -GNinja \
       -DPATH_TO_LLVM=$LLVM_ROOT \
       -DCMAKE_BUILD_TYPE=Release \
       -DCMAKE_INSTALL_PREFIX=$LLVM_ROOT \
       .. \
    && ninja \
    && ninja install
```

The build fail at last build step, i get 20+ linker errors:
```
[116/117] Linking CXX executable tools/internal/mutator-validator/mull-mutator-validator
FAILED: tools/internal/mutator-validator/mull-mutator-validator 
[...]

PR #758, opened 2020-09-25
Add a short hint regarding #657 to HackingOnMull.rst

ea1111:
I just run into the same issue that was described in #657 and the solution mentioned there worked for me. So I added a short hint to HackingOnMull.rst.


AlexDenisov (approved):

AlexDenisov:
Thanks a lot @ea1111!

PR #759, opened 2020-10-01
cmake: Refactor test handling

gebart:
Use CTest and follow recommended practices.

Took some suggestions from this piece: https://cliutils.gitlab.io/modern-cmake/chapters/testing.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,10 @@ project(Mull
   LANGUAGES C CXX
 )
 
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+  include(CTest)
+endif()
+
 # -DRECORD_TIMING=ON
 if (RECORD_TIMING)
   set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "\"${CMAKE_COMMAND}\" -E time")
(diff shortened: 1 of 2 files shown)


gebart:
Seems like a random network failure on the 16.04/ LLVM 8.0 CI build.

AlexDenisov:
Could you add a few words on the benefits of the CTest? I tried to use it couple of times but I didn't get it :)

gebart:
In essence, this PR changes two things:
- Add a setting `BUILD_TESTING` which turns on/off the compilation of the unit tests, instead of having them always configured but added as `EXCLUDE_FROM_ALL`. This comes from `include(CTest)` in CMakeLists.txt.
- Use `ctest` to run tests when run from the makefile or other build environment. The main benefit is that it logs all stdout and stderr output from the tests, and it can run tests in parallel if we have multiple test targets.

Right now there is only a single unit test binary and only a single test registered in CMake, so nothing will be parallelized, but we could use `gtest_discover_tests()` to allow the test cases in that single binary to be run in parallel if we bump the CMake requirement to version 3.10. I felt that this was better left for a separate discussion after this PR has been merged though.

gebart:
The `if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)` part is to avoid our unit tests being built and run if someone were to include mull via `add_subdirectory` inside their own project cmake hierarchy. I wrote this only as a means to follow best practices, I currently don't have a project where I want to build mull as a subdirectory.

AlexDenisov (approved):

PR #760, opened 2020-10-01
CompilationDatabase refactor

gebart:
This started out as a refactor of the compilation database handling to make the code a bit more readable and testable in order to be able to replace the unescape implementation which was determined via profiling to be extremely CPU costly at runtime. However, after doing the refactoring I realized that we could simply use the compilation database parser in the clang tooling instead of reimplementing our own, so the final commit replaces much of the parsing and handling with calls to clang code. These functions are part of the public API as far as I can tell and should be available if everything that is already needed to build mull is available.

The change away from using `std::regex_replace` in the unescape function reduces the time to load the compilation database in one large project by 25 seconds on a mull run that takes 50 seconds in total for a particular test binary.


gebart:
Seems like the clang tooling API around JSONCompilationDatabase was modified for the Clang 4.0 release.
Commit: llvm/llvm-project@9e60a2ad734e1d598f93e7148c6339795b57359c

gebart:
@AlexDenisov what is your opinion on this wrt Clang 3.9. Some alternatives:
- Add a new wrapper function in LLVMCompatibility for JSONCompilationDatabase to set the argument `syntax=autodetect` when LLVM version >= 4.0
- Drop the Clang tooling commits, (keep f44bee1 and earlier commits)
- Drop LLVM 3.9 support

gebart:
(Force-pushed to trigger a CI build on the commit before the Clang tooling commits)

AlexDenisov:
> Drop LLVM 3.9 support

I like this one :)
I'm planning to bring Mull to the LLVM 10 over the next few days, and then we can drop the 3.9. if it's a strong blocker, then we can drop it earlier.

gebart:
pushed the Clang tooling commits back into this PR.
This PR now depends on a PR dropping 3.9 support.

gebart:
depends on #762

gebart:
Rebased to incorporate C++17 bump

AlexDenisov (approved):

PR #762, opened 2020-10-03
Drop 'old' LLVM versions

AlexDenisov:


AlexDenisov:
This should enable us to switch to C++17 as suggested by @gebart in the offline (slack) discussion.
cc @stanislaw

gebart:
Nice that you were able to remove some of the abstractions that were no longer needed :+1:

gebart:
Did you pick version 6.0 as the lowest version because that is what is available in Ubuntu 18.04 LTS and Debian stable repos?
From a user perspective I think it makes sense to support what is easily installable in the most common "stable" distribution channels (Ubuntu, Debian, Homebrew, maybe Centos)

AlexDenisov:
@gebart that's the first one that compiled with `-std=c++17`, at least that was the case with `libirm`.

PR #764, opened 2020-10-04
Turn on C++17

AlexDenisov:


CMakeLists.txt
```diff
@@ -147,7 +150,7 @@ endif()
 
 set(MULL_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
-set (MULL_CXX_FLAGS "-std=c++14 -fPIC -fvisibility-inlines-hidden -fno-exceptions -Wunused -Wall -Werror")
+set (MULL_CXX_FLAGS "-fPIC -fvisibility-inlines-hidden -fno-exceptions -Wunused -Wall -Werror")
```
gebart, line 153: `-fvisibility-inlines-hidden` can also be set in cmake via  [`CMAKE_VISIBILITY_INLINES_HIDDEN`](https://cmake.org/cmake/help/latest/variable/CMAKE_VISIBILITY_INLINES_HIDDEN.html)
-fPIC also has a cmake property but that one is usually set automatically when it is needed.

gebart (commented):
Some suggestions inline

CMakeLists.txt
AlexDenisov: This one conflicts with the json library (just shows some warnings).

AlexDenisov:
Our CMake setup certainly needs some love :)

AlexDenisov:
I guess Ubuntu 16 has some old version of libc++, at which point I'm ready to just drop 16.04 and move on. I didn't see any ubuntu16 downloads for several months.
If we want to support that version, then we can have a separate discussion.

PR #767, opened 2020-10-04
Drop Ubuntu 16 build on CI

AlexDenisov:


AlexDenisov:
I canceled CI builds, no sense in running them against this PR.

PR #772, opened 2020-10-13
Add timeout config options

AlexDenisov:


gebart (changes requested):
Nice, this is something that I have use of for some of my test cases. There is already a default timeout constant which could be used as the default value in cli options too, to avoid repeating the same literal twice.

gebart (approved):
Looks good!

PR #779, opened 2020-11-18
cmake: Search for SQLite3 during configuration phase

gebart:
Signal an error immediately in CMake instead of failing during
compilation if SQLite is not installed.


AlexDenisov:
@gebart you'd also need to bump the CMake version here https://github.com/mull-project/mull/blob/master/infrastructure/ubuntu-playbook.yaml#L45

PR #793, opened 2020-12-16
Control mutations with ENV vars instead of global pointers

AlexDenisov:
Instead of doing magical shenanigans with pointers, there is now a slightly better approach.
Here how it was before:

```c
void *pointer = NULL;
void f_original() {}
void f_mutant1() {}
void f_mutant2() {}

void f() {
  pointer(); // where pointer value is set to f_original, f_mutant1, or f_mutant2 at runtime via JIT tricks
}
```

and here is how it's done now:

```c
void f_original() {}
void f_mutant1() {}
void f_mutant2() {}

void f() {
  void *pointer = NULL;
  if ("enable_mutant1") { pointer =  f_mutant1; }
  else if ("enable_mutant2") { pointer =  f_mutant2; }
  else { pointer = f_original; }
  pointer();
}
```
Technically, the behavior is still the same, but it is now a bit easier to manipulate mutants without relying on JIT engine.


AlexDenisov:
CI on macOS fails because the run against OpenSSL takes much more time.
I did some benchmarking on my machine and on average the master branch finishes in ~100 seconds, while this change slows everything down to around ~600 seconds.

AlexDenisov:
Just hacked an attempt to implement the same behavior, but with jump tables (huge switch) instead of a long if/else chain.
Performance is pretty good (also around ~100s), but there are a few gotchas:
 - we cannot ensure support for mutations in static constructors (because of the initialization order fiasco)
 - we can only support one mutation per run
 - it requires the emission of an additional function (to initialize `currentMutant`)
 - ^ adds a (tiny) overhead during the initialization

(Perhaps, it makes sense to implement both behaviors under a switch? Anyhow, I'll get back to it later).

(The code is here, but it's only for "historical" reference (seriously, don't look at it): https://github.com/mull-project/mull/compare/drop-trampoline-pointers-jump-tables?expand=1)

PR #795, opened 2020-12-19
Docs: fix link to community chat

ligurio:


AlexDenisov:
Good catch, I updated readme but forgot about this link.

PR #796, opened 2020-12-19
Docs: add a page with a list of key features

ligurio:
It would be helpful for newcomers to describe Mull key features in a documentation.

### DIFF
--- a/docs/Features.rst
+++ b/docs/Features.rst
@@ -1,0 +1,9 @@
+Features
+========
+
+Key features are:
+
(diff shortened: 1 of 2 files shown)


AlexDenisov:
Thank you @ligurio, I like the list. I have an impression that the testing framework support is a bit misleading (i.e. Mull supports any framework, though there are nice features when we know specifics of a test framework), but I'll write a separate tech note to clarify what a testing framework support means :)

PR #806, opened 2020-12-29
Upstreaming JIT removal

AlexDenisov:
These changes are needed to drop JIT, but they do not break existing master


AlexDenisov:
This is the second piece, the first one was #793

stanislaw (commented):
👍

PR #808, opened 2020-12-29
Emit normal executable instead of using JIT.

AlexDenisov:
This is a work in progress, but I'd like to get it into the master before completeness otherwise the final PR will be a huge mess.
The primary goal at the moment is to make CI green for this PR.
Any comments/suggestions/concerns are very welcome.


AlexDenisov:
Not ready for review yet.

lib/Driver.cpp
```diff
@@ -93,39 +94,44 @@ std::vector<MutationPoint *> Driver::findMutationPoints(vector<Test> &tests) {
...
+                       << "\n";
+        failureMessage << "status: ";
+        failureMessage << result.getStatusAsString() << "\n";
+        failureMessage << "stdout: '";
+        failureMessage << result.stdoutOutput << "'\n";
+        failureMessage << "stderr: '";
+        failureMessage << result.stderrOutput << "'\n";
+        diagnostics.warning(failureMessage.str());
+      }
+    });
```
AlexDenisov, line 124: This is here just to test the basic compatibility between reproc and our own implementation. This piece will be wrapped and removed in one of the next iterations.

lib/Driver.cpp
```diff
@@ -315,3 +318,84 @@ Driver::Driver(Diagnostics &diagnostics, const Configuration &config, const Proc
...
+          for (llvm::Function &function : *bitcode->getModule()) {
+            functionsUnderTest.emplace_back(&function, &test, 1);
+          }
+        }
+      }
+    }
+  });
+
+  return functionsUnderTest;
+}
```
AlexDenisov, line 401: This should obviously not be here, but I'll extract and move it away as soon as the dead code is gone.

lib/Toolchain/Compiler.cpp
```diff
@@ -38,3 +45,76 @@ OwningBinary<ObjectFile> Compiler::compileModule(Module *module, TargetMachine &
...
+    llvm_compat::writeBitcodeToFile(*bitcode.getModule(), bcStream);
+    diagnostics.debug("Emitted object file: "s + resultPath.c_str() + " for bitcode file: "s +
+                      bcPath.c_str());
+  }
+
+  pass.run(*bitcode.getModule());
+  dest.flush();
+
+  return resultPath.str().str();
+}
```
AlexDenisov, line 120: This piece uses LLVM's API to generate an object file. It may make sense to actually just rely on clang/clang++ in case we have a compilation database. Otherwise, we can wrongly emit 'slightly' incorrect/unexpected object files due to different defaults.

lib/Toolchain/Linker.cpp
```diff
@@ -0,0 +1,49 @@
...
+    message << "Cannot link program\n";
+    message << "command: " << command << "\n";
+    message << "stdout: " << result.stdoutOutput << "\n";
+    message << "stderr: " << result.stderrOutput << "\n";
+    diagnostics.error(message.str());
+  }
+  diagnostics.debug("Link command: "s + command);
+  diagnostics.debug("Compiled executable: "s + resultPath.c_str());
+  return resultPath.str().str();
+}
```
AlexDenisov, line 49: Here, we will eventually face the problem of 'too many arguments'. Not sure what would be the right solution, but I'll postpone it till the issue actually appears.

lib/Toolchain/Runner.cpp
```diff
@@ -0,0 +1,83 @@
...
+  if (!captureOutput) {
+    reproc::drain(process, reproc::sink::null, reproc::sink::null);
+    return std::make_pair("", "");
+  }
+  std::string out;
+  reproc::sink::string outSink(out);
+  std::string err;
+  reproc::sink::string errSink(err);
+  // Ignoring errors here
+  reproc::drain(process, outSink, errSink);
```
AlexDenisov, line 24: Something is not quiet right here, so it needs a revision in the future.

tests/DriverTests.cpp
```diff
@@ -48,13 +48,14 @@ TEST(Driver, RunningWithNoTests) {
   ASSERT_EQ(0u, result->getMutationResults().size());
 }
 
-TEST(Driver, customTest_withDynamicLibraries) {
+TEST(Driver, DISABLED_customTest_withDynamicLibraries) {
```
AlexDenisov, line 51: These must migrate to LIT, but I was too lazy to actually finish these ones as well.

tools/mull-cxx/CLIOptions.cpp
```diff
@@ -93,7 +93,7 @@ opt<std::string> tool::Linker(
     desc("Linker program"),
     value_desc("string"),
     Optional,
-    init("clang"),
```
AlexDenisov, line 96: This is an unsafe default: only works sometimes by accident.

Issue #636, opened 2019-12-05
Feature request: mull line coverage


Warwolt:
(Writing this feature request after a twitter conversation with Alex Denisov)

## User story:
_"As a developer looking to verify the completeness of my unit tests, I want Mull to be able to produce the line coverage of the code under test, so that I can be sure that the 100% mutant kill rate is based on as many mutants as possible for a given source file."_

## Motivation
The basic idea is that, for a given source file `module.c` with a test suite `module_test.cc` there should be some set **M** of all mutants possible to generate with Mull. If the line coverage of `module_test.cc` is less than 100%, it is possible that the set of actually generated mutants **M´** form a _strict_ subset of the set of **M**, that is **M'** ⊂ **M**.

The kill rating is based on the number of kills to the total number of mutants, with the goal of getting a 100% kill rate. When the set of mutants being trialed with the test suite is the subset **M'**, then the 100% kill rate represents less behavioural coverage than the 100% kill rate for the set **M**. 

The goal of mutation testing is to measure to the extent that a test suite enforces the behaviour of the code under test, and the ability for the mutation tool to perform this task is strongly related to the line coverage of the test suite.

As an illustrative example, imagine that `module.c` contains 10 functions each with 10 possible mutants, but only 2 of the functions are fully line covered by `module_test.cc`. Even with a 100% mutant kill rate, we have only covered 20% of the possible behaviour that Mull would be able to detect, and therefore still have improvements to make that aren't reported. 

My feature request therefore is to add a line coverage metric to Mull during mutation testing, and report this along with the unkilled mutants, to guide developers in gauging how much more work needs before a test suite can be considered done. 

Knowing that there is a 100% line coverage and a 100% mutant kill rate would give _strong_ arguments for the validity of the test suite being measured, and so would add I believe quite a lot of value to the already good Mull mutation tool.

chgans:
I do exactly that, i build and run my tests with code coverage, and then I build and run them again with mutation. I end up with 2 metrics. 
Are you asking for mull to be compatible with code coverage? so that (in my case) and only need 1 build and 1 run?

Warwolt:
Yes essentially, I've seen other mutation testing tools able to tell the amount of coverage in the test report and imply that there isn't a true 100% kill rate until all code is covered.

Of course running a coverage separately is an easy work around, but this would be a real Nice-To-Have.

AlexDenisov:
[...]

PR #831, opened 2021-01-09
Add support for not covered mutants

AlexDenisov:
By default Mull generates mutations for every function possible. There is an option (`-coverage-info`) to use the coverage information obtained from Clang (see the [docs](https://mull.readthedocs.io/en/latest/ControlMutationsTutorial.html#code-coverage) for more details) to not mutate unreachable code.

Without coverage information, Mull wastes CPU time by checking mutants that cannot be killed. With the coverage information Mull gives misleading reports saying that the mutation coverage is 100% even if only 10% of code was mutated.

This patch adds an option (`-include-not-covered`) to also include not covered mutants in the report even when code coverage is provided. The not-covered mutants will show up in the reports, but Mull won't actually mutate anything and won't run tests against not covered mutants.

This patch partially addresses #636.

P.S. This is up for review, but I'll do more testing soon as I may have missed some edge cases.

### DIFF
--- a/docs/generated/CLIOptions.rst
+++ b/docs/generated/CLIOptions.rst
@@ -44,6 +44,8 @@
 
 --coverage-info string		Path to the coverage info file (LLVM's profdata)
 
+--include-not-covered		Include (but do not run) not covered mutants. Disabled by default
+
 --include-path regex		File/directory paths to whitelist (supports regex)
 
 --exclude-path regex		File/directory paths to ignore (supports regex)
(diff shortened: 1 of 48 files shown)


stanislaw (approved):

docs/generated/CLIOptions.rst
```diff
@@ -44,6 +44,8 @@
 
 --coverage-info string		Path to the coverage info file (LLVM's profdata)
 
+--include-not-covered		Include (but do not run) not covered mutants. Disabled by default
```
Warwolt, line 47: nit: should this not say "non-covered" mutants? "not covered" sounds a little strange in plural.

AlexDenisov: Yeah, it makes sense. There are some contexts (like "Include (but do not run) not covered mutants." above) where "not" is more appropriate, imo. So I think it's better to stick to one form to avoid any confusion.

PR #833, opened 2021-02-11
Incremental mutation testing using GitDiffFilter

stanislaw:
Mull runs ``git diff -U0`` under the hood which means getting diff with zero lines of context.

The implementation is inspired by the approach used by git-clang-format:

https://opensource.apple.com/source/clang/clang-800.0.38/src/tools/clang/tools/clang-format/git-clang-format.auto.html

---

TODO:

- [x] Agree on the CLI interface
- [x] Make sure that the full paths in the debug information are not messed up
- [x] Documentation

### DIFF
--- a/lib/Filters/GitDiffFilter.cpp
+++ b/lib/Filters/GitDiffFilter.cpp
@@ -1,0 +1,82 @@
+#include "mull/Filters/GitDiffFilter.h"
+
+#include "mull/Diagnostics/Diagnostics.h"
+#include "mull/SourceLocation.h"
+#include "mull/Toolchain/Runner.h"
+
+#include "llvm/IR/InstIterator.h"
+#include <llvm/IR/DebugInfoMetadata.h>
+#include <llvm/IR/DebugLoc.h>
+#include <llvm/IR/Function.h>
+#include <llvm/Support/FileSystem.h>
+
+#include <iostream>
+#include <regex>
+#include <sstream>
+#include <unistd.h>
+
+using namespace mull;
+
+GitDiffFilter *GitDiffFilter::createFromGitDiff(Diagnostics &diagnostics,
+                                                const std::string &gitProjectRoot,
+                                                const std::string &gitDiffBranch) {
+  mull::GitDiffReader gitDiffReader(diagnostics, gitProjectRoot);
+  mull::GitDiffInfo gitDiffInfo = gitDiffReader.readGitDiff(gitDiffBranch);
+  return new GitDiffFilter(diagnostics, gitDiffInfo);
+}
+
+GitDiffFilter::GitDiffFilter(Diagnostics &diagnostics, GitDiffInfo gitDiffInfo)
+    : diagnostics(diagnostics), gitDiffInfo(gitDiffInfo) {}
+
+std::string GitDiffFilter::name() {
+  return "Git Diff";
+}
+
+bool GitDiffFilter::shouldSkip(llvm::Instruction *instruction) const {
+  SourceLocation sourceLocation = SourceLocation::locationFromInstruction(instruction);
+  if (sourceLocation.isNull()) {
+    return true;
+  }
+
+  /// If no diff, then filtering out.
+  if (gitDiffInfo.size() == 0) {
+    std::stringstream debugMessage;
+    debugMessage << "GitDiffFilter: git diff is empty. Skipping instruction: ";
+    debugMessage << sourceLocation.filePath << ":";
+    debugMessage << sourceLocation.line << ":" << sourceLocation.column;
+    diagnostics.debug(debugMessage.str());
+    return true;
+  }
+
+  /// If file is not in the diff, then filtering out.
+  if (gitDiffInfo.count(sourceLocation.filePath) == 0) {
+    std::stringstream debugMessage;
+    debugMessage << "GitDiffFilter: the file is not present in the git diff. ";
+    debugMessage << "Skipping instruction: ";
+    debugMessage << sourceLocation.filePath << ":";
+    debugMessage << sourceLocation.line << ":" << sourceLocation.column;
+    diagnostics.debug(debugMessage.str());
+    return true;
+  }
+
+  GitDiffSourceFileRanges ranges = gitDiffInfo.at(sourceLocation.filePath);
+  for (auto &range : ranges) {
+    int rangeEnd = range.first + range.second - 1;
+    if (range.first <= sourceLocation.line && sourceLocation.line <= rangeEnd) {
+      std::stringstream debugMessage;
+      debugMessage << "GitDiffFilter: whitelisting instruction: ";
+      debugMessage << sourceLocation.filePath << ":";
+      debugMessage << sourceLocation.line << ":" << sourceLocation.column;
+      diagnostics.debug(debugMessage.str());
+      return false;
+    }
+  }
+
+  std::stringstream debugMessage;
+  debugMessage << "GitDiffFilter: skipping instruction: ";
+  debugMessage << sourceLocation.filePath << ":";
+  debugMessage << sourceLocation.line << ":" << sourceLocation.column;
+  diagnostics.debug(debugMessage.str());
+
+  return true;
+}
--- a/lib/Filters/GitDiffReader.cpp
+++ b/lib/Filters/GitDiffReader.cpp
@@ -1,0 +1,76 @@
+#include "mull/Filters/GitDiffReader.h"
+
+#include "mull/Diagnostics/Diagnostics.h"
+#include <mull/Path.h>
+#include "mull/Toolchain/Runner.h"
+
+#include <iostream>
+#include <regex>
+#include <sstream>
+
+using namespace mull;
+
+GitDiffReader::GitDiffReader(Diagnostics &diagnostics, const std::string gitRepoPath)
+    : diagnostics(diagnostics), gitRepoPath(gitRepoPath) {}
+
+GitDiffInfo GitDiffReader::readGitDiff(const std::string &gitBranch) {
+  /// The implementation is borrowed from the git-clang-format Python tool.
+  /// https://opensource.apple.com/source/clang/clang-800.0.38/src/tools/clang/tools/clang-format/git-clang-format.auto.html
+  Runner runner(diagnostics);
+
+  std::vector<std::string> args = { "diff", "-U0", gitBranch };
+
+  ExecutionResult result =
+      runner.runProgram("git", args, {}, 5000, true, gitRepoPath);
+  if (result.exitStatus != 0) {
+    diagnostics.warning(
+        std::string("GitDiffReader: cannot get git diff information. Received output: ") +
+        result.stderrOutput);
+    return GitDiffInfo();
+  }
+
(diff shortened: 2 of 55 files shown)


lib/Filters/GitDiffFilter.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <llvm/Support/FileSystem.h>
+
+#include <iostream>
+#include <regex>
+#include <sstream>
+#include <unistd.h>
+
+using namespace mull;
+
+static std::string getCwd() {
```
stanislaw, line 19: Do we already have this kind of utils somewhere?

lib/Filters/GitDiffFilter.cpp
```diff
@@ -0,0 +1,97 @@
...
+GitDiffFilter::GitDiffFilter(Diagnostics &diagnostics)
+    : diagnostics(diagnostics), gitDiffInfoStorage() {}
+
+std::string GitDiffFilter::name() {
+  return "Git Diff";
+}
+
+void GitDiffFilter::activateAgainstGitBranch(const std::string &gitDiffBranch) {
+  /// Assuming that Mull is run from the project's directory.
+  const std::string gitRootPath = getCwd();
```
stanislaw, line 36: I am not sure about this one: it is either communicating this in the docs that Mull has to run from the project's Git root or provide an extra argument that specifies the path to Git root.

lib/Filters/GitDiffFilter.cpp
```diff
@@ -0,0 +1,97 @@
...
+  auto debugInfo = llvm::dyn_cast<llvm::DISubprogram>(function->getMetadata(0));
+
+  std::string unitDirectory;
+  std::string unitFilePath;
+  if (llvm::DICompileUnit *unit = debugInfo->getUnit()) {
+    unitDirectory = unit->getDirectory();
+    unitFilePath = unit->getFilename();
+  }
+  assert(unitFilePath[0] == '/');
+  assert(llvm::sys::fs::exists(unitFilePath));
```
stanislaw, line 54: This might work in the LIT tests because there we are compiling from the `cd /` root. I would like to check this when everything is compiled not from the root.

lib/Filters/GitDiffFilter.cpp
```diff
@@ -0,0 +1,97 @@
...
+}
+
+bool GitDiffFilter::shouldSkipSourceFunction(const std::string &fullPath, const int beginLine,
+                                             const int endLine) {
+  /// No diff, not filtering.
+  if (gitDiffInfoStorage.size() == 0) {
+    return false;
+  }
+
+  /// If there is an intersection between the function's [begin, end] range and
```
stanislaw, line 85: Do you like this approach? The whole function gets included if at this one range intersects with it.

tests-lit/Makefile
```diff
@@ -23,7 +23,6 @@ test: ## Run integration tests
 CLEAN_FIND_CMD=find \
 	. \
 	-type f \( \
```
stanislaw, line 25: We have a precompiled `llvm-11-macos-bitcode-sample.exe` file (04-bitcode-llvm-version-mismatch) that was deleted every time this command was run. Having `Output` folder included as `-name 'Output'` ensures that we delete all the `.exe` files from the `Output`.

tools/mull-cxx/CLIOptions.cpp
```diff
@@ -139,6 +139,13 @@ opt<bool> tool::EnableAST(
   llvm::cl::desc("Enable \"white\" AST search (disabled by default)"),
   llvm::cl::cat(MullCXXCategory), llvm::cl::init(false));
 
+opt<std::string> tool::GitDiff(
+    "git-diff",
```
stanislaw, line 143: Would it make sense to call this `experimental-git-diff` for now to indicate alpha/beta testing phase?

tools/mull-cxx/mull-cxx.cpp
```diff
@@ -211,6 +212,12 @@ int main(int argc, char **argv) {
     filePathFilter->include(regex);
   }
 
+  if (tool::GitDiff.getValue().size() > 0) {
```
stanislaw, line 215: I might have missed the right way of testing the presence of an optional CLI string argument. Do you know a better way?

lib/Filters/GitDiffReader.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  std::string currentFileName;
+
+  while (std::getline(diffStream, currentLine, '\n')) {
+    std::regex lineRegex("^\\+\\+\\+\\ [^/]+/(.*)");
+    std::regex rangeRegex("^@@ -[0-9,]+ \\+(\\d+)(,(\\d+))?");
+    std::smatch matches;
+
+    if (std::regex_search(currentLine, matches, lineRegex)) {
+      // TODO: What is the orthodox way of joining path components?
```
stanislaw, line 56: Pretty much what TODO says.

lib/Filters/GitDiffFilter.cpp
stanislaw: Just checked: C++17 has `#include <filesystem>` and this API:

```cpp
    namespace fs = std::filesystem;
    ...
    fs::path dir ("/tmp");
    fs::path file ("foo.txt");
    fs::path full_path = dir / file;
    std::cout << full_path << std::endl;
```

But on my macOS 10.14 is says this is unavailable until 10.15.

lib/Filters/GitDiffFilter.cpp
AlexDenisov: We better use this one [`llvm::sys::fs::current_directory`](https://llvm.org/doxygen/namespacellvm_1_1sys_1_1fs.html#aec50dfe28ae22b8ff45f02ca193b7bf7)

lib/Filters/GitDiffFilter.cpp
AlexDenisov: Let's just have it in the docs for the beginning

lib/Filters/GitDiffFilter.cpp
AlexDenisov: You can re-use this part. I suppose: https://github.com/mull-project/mull/blob/master/lib/SourceLocation.cpp#L52

re: `std::filesystem` so far I don't see a reason to use it, we can just rely on the LLVM's fs/path modules and call it a day :)

lib/Filters/GitDiffFilter.cpp
AlexDenisov: It's a good start for sure! Though, I think it makes more sense to actually filter out individual mutations based on the diff range.
But let's start with this approach and then implement the rest.

lib/Filters/GitDiffReader.cpp
AlexDenisov: We have couple more places where "/" is hardcoded, so whatever :)

tools/mull-cxx/CLIOptions.cpp
AlexDenisov: I don't think so, though I would pick another name, like `git-diff-ref`or something similar.

tools/mull-cxx/mull-cxx.cpp
AlexDenisov: `hasValue()` maybe? Anywhow, it's recommended to use `empty()` instead of `size() == 0`

AlexDenisov (approved):
You'd also need to run `build-docs` to regenerate CLI options doc (to include the `git-diff` option)

tools/mull-cxx/mull-cxx.cpp
stanislaw: `hasValue` returns true when there is no option provided. I have switched to `empty()`.

lib/Filters/GitDiffReader.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+GitDiffInfo GitDiffReader::readGitDiff(const std::string &gitBranch) {
+  /// The implementation is borrowed from the git-clang-format Python tool.
+  /// https://opensource.apple.com/source/clang/clang-800.0.38/src/tools/clang/tools/clang-format/git-clang-format.auto.html
+  Runner runner(diagnostics);
+
+  std::vector<std::string> args = { "diff", "-U0", gitBranch };
+  std::vector<std::string> env = {};
+
+  ExecutionResult result = runner.runProgram("git", args, env, 1000, true);
```
AlexDenisov, line 24: 1 second seems to be too optimistic. Shall we make it configurable as well?

AlexDenisov: Also, don't you want to `cd` into `gitRepoPath` first? Since we require this piece it makes sense to not require running Mull from the git repo directory.

tools/mull-cxx/CLIOptions.cpp
```diff
@@ -139,6 +139,20 @@ opt<bool> tool::EnableAST(
   llvm::cl::desc("Enable \"white\" AST search (disabled by default)"),
   llvm::cl::cat(MullCXXCategory), llvm::cl::init(false));
 
+opt<std::string> tool::GitDiffRef(
+    "git-diff-ref",
+    desc("Git branch to run diff against (enables incremental testing)"),
+    Optional,
+    value_desc("git branch name"),
```
AlexDenisov, line 146: Could also be a tag or a commit hash, right?

AlexDenisov (approved):

lib/Filters/GitDiffReader.cpp
stanislaw: How about simply increasing this one to `5000`? It is a one-time action so should be safe enough.

stanislaw: > Also, don't you want to cd into gitRepoPath first? 

Good catch. I will do it.

tools/mull-cxx/CLIOptions.cpp
stanislaw: I have checked `git diff --help` just now:

```
NAME
       git-diff - Show changes between commits, commit and working tree, etc

SYNOPSIS
       git diff [<options>] [<commit>] [--] [<path>...]
       git diff [<options>] --cached [<commit>] [--] [<path>...]
       git diff [<options>] <commit> <commit> [--] [<path>...]
       git diff [<options>] <blob> <blob>
       git diff [<options>] --no-index [--] <path> <path>
```

So, I went with `<commit>`.

lib/Toolchain/Runner.cpp
```diff
@@ -82,3 +82,54 @@ ExecutionResult Runner::runProgram(const std::string &program,
 
   return result;
 }
+
+ExecutionResult Runner::runProgramFromWorkingDirectory(const std::string &program,
```
stanislaw, line 86: I avoided extending the existing `runProgram` method with an optional `working directory` argument and simply created another focused method. Also, the existing method has code that does some manipulation with the environment, so I thought it made sense to remove it from the new `runProgramFromWorkingDirectory`. (Hope this is ok.)

stanislaw: I have decided to merge them into one 😃

AlexDenisov (approved):

PR #834, opened 2021-02-24
Copy call inst attributions and calling convention

kateinoigakukun:
Thanks for developing this awesome software!
I'm trying to use this for XCTest and Swift. Here is my WIP repository: https://github.com/kateinoigakukun/mull-xctest

Then Swift uses its own special calling convention and attributes, so call inst in trampoline func should also have those attributes.


AlexDenisov (approved):

AlexDenisov:
Thanks, @kateinoigakukun! It's a nice one, we should be doing this regardless of the Swift compatibility!

Issue #778, opened 2020-11-18
Support tests written on interpreted languages


ligurio:
In a most C/C++ projects only small part of tests written on C/C++. Other parts are implemented on Interpreted languages like Python, Perl etc, because they are cheaper in implementation and maintenance. As far as I understand Mull supports only tests written on C/C++. It means that it cannot kill all produced mutants even there are exists tests that can do it, because Mull don't run them.

It would be good to add ability to execute all possible tests exist in a project. Otherwise mutation testing using Mull is imperfect.

gebart:
I think this is a good suggestion. To implement this I assume that we would have to be able to create a mutated standalone binary that can be run in the same way as the original binary. Not sure how to take this from the current binary in memory, but one potential solution would involve calling a normal linker to be able to create a freestanding executable. Given the wide range of build systems and how they are invoked I think the easiest way might even be to provide hooks where Mull calls an external script/program for linking and another external program for executing the tests. That would allow for full flexibility regarding how test binaries should be built and how they are invoked.

Some open questions that I see around this approach:

- Does it make sense to implement something like this in mull-cxx, or is this out of scope for this tool?
- Are there any other mutation testing products (for C++ code) which already have this feature?
- Running multiple mutations in parallel may pose an obstacle when built this way. Who should be responsible to ensure that each parallel run has its own working area, Mull or the external scripts?
- Should each mutation have its own directory, or is it sufficient to have separate working directories per worker thread so that parallel runs are not stepping on each others' files?
- Who cleans up object files and other temporaries after a mutation run?

AlexDenisov:
I'd love to see Mull supporting this use case. There are, however, obstacles.

Here are some technical details.

1. As @gebart mentioned, this approach requires Mull to generate a standalone executable. We can get all the objects files from a program by compiling embedded bitcode to machine code (pretty much the same way we do now with the JIT engine). However, we'd also need to get linker flags from a user so that we can emit correctly linked executable. An alternative approach - store all the object files in some folder and ask a user to link them manually (e.g. `clang *.o -o the-tests`).

2. Mull relies on the code coverage info to generate fewer mutations. Currently, Mull gathers coverage by running tests before generating any mutants. In the case of a standalone executable, we can no longer run the tests via JIT so we have to add support for external coverage information (e.g., gcov, lcov, etc.). In this case, we will also lose the call-tree feature which is useful when combined with a specific test framework.

[...]

PR #835, opened 2021-02-25
[Refactoring] Remove Mutant dependency from reporters

kateinoigakukun:
I'm trying to separate modification and execution processes to support XCTest harness. 
Now reporters depend on `Mutant` type, and it makes it hard to separate processes even though they only use source location and mutator kind.

This PR changes to pass only essential things.

### DIFF
--- a/include/mull/JunkDetection/CXX/ASTStorage.h
+++ b/include/mull/JunkDetection/CXX/ASTStorage.h
@@ -45,7 +45,7 @@ public:
              const std::string &cxxCompilationFlags,
              const std::map<std::string, std::string> &bitcodeCompilationFlags);
 
-  ThreadSafeASTUnit *findAST(const MutationPoint *point);
+  ThreadSafeASTUnit *findAST(const SourceLocation &sourceLocation);
   ThreadSafeASTUnit *findAST(const std::string &sourceFile);
 
   void setAST(const std::string &sourceFile, std::unique_ptr<ThreadSafeASTUnit> astUnit);
--- a/include/mull/Reporters/ASTSourceInfoProvider.h
+++ b/include/mull/Reporters/ASTSourceInfoProvider.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "mull/Mutators/MutatorKind.h"
+
 namespace mull {
 
 struct MutationPointSourceInfo {
@@ -12,11 +14,13 @@ struct MutationPointSourceInfo {
 class ASTStorage;
 class MutationPoint;
 class Diagnostics;
+struct SourceLocation;
 
 class SourceInfoProvider {
 public:
   virtual MutationPointSourceInfo getSourceInfo(Diagnostics &diagnostics,
-                                                MutationPoint *mutationPoint) = 0;
+                                                const SourceLocation &sourceLocation,
+                                                MutatorKind mutatorKind) = 0;
 };
 
 class ASTSourceInfoProvider : public SourceInfoProvider {
@@ -24,7 +28,8 @@ public:
   ~ASTSourceInfoProvider() = default;
   explicit ASTSourceInfoProvider(ASTStorage &astStorage);
   MutationPointSourceInfo getSourceInfo(Diagnostics &diagnostics,
-                                        MutationPoint *mutationPoint) override;
+                                        const SourceLocation &sourceLocation,
+                                        MutatorKind mutatorKind) override;
 
 private:
   ASTStorage &astStorage;
--- a/lib/JunkDetection/CXX/ASTStorage.cpp
+++ b/lib/JunkDetection/CXX/ASTStorage.cpp
@@ -175,17 +175,16 @@ ASTStorage::ASTStorage(Diagnostics &diagnostics, const std::string &cxxCompilati
           diagnostics, cxxCompilationDatabasePath, cxxCompilationFlags, bitcodeCompilationFlags)),
       mutations(diagnostics) {}
 
-ThreadSafeASTUnit *ASTStorage::findAST(const MutationPoint *point) {
-  assert(point);
-  assert(!point->getSourceLocation().isNull() && "Missing debug information?");
+ThreadSafeASTUnit *ASTStorage::findAST(const SourceLocation &sourceLocation) {
+  assert(!sourceLocation.isNull() && "Missing debug information?");
 
-  const std::string &sourceFile = point->getSourceLocation().unitFilePath;
+  const std::string &sourceFile = sourceLocation.unitFilePath;
   if (llvm::sys::fs::exists(sourceFile)) {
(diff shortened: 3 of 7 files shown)


AlexDenisov:
@kateinoigakukun could you please explain the reason for this change? I didn't get it.

kateinoigakukun:
@AlexDenisov 
Sorry for my too short explanation. 

Let me explain my experimental approach for Swift and XCTest.

First, I ended up that it's a little bit hard approach to embed LLVM bitcode in the **final executable file** for Swift because:

1. Swift has some runtime libraries and some of them are linked statically. But they doesn't have bitcode information, so ld64 drops bitcode at all.
2.  And those runtime libraries needs to be linked when re-linking object files in mull pipeline, so users have to pass linker flags manually. 
3. (for iOS environments) Executable files must be signed, so we need to re-sign after modification.

So I'm trying to modify LLVM instructions while linking phase to avoid modifying the final executable file.

Here is my rough blueprint for the testing pipeline.


1. Compile source code to object file with bitcode
swiftc main.swift 

2. Link object files with mull-ld
mull-ld main.o -lswiftCore -lswiftCompatibility51
   1. mull-ld extracts bitcode from object files
   2. Modify LLVM instructions for mutation testing.
   3. Embed mutant identifiers in a special section in object files.
   3. Then compiles them into object files
   4. Finally delegate compiled object files to ld64 with given options.

3. Execute the produced binary with mull-xctest command for mutation testing.
    1. Extract mutant identifiers from the special section from executable file.
    2. Execute `MutantExecutionTask` and reports results.


As this step shows, `mull-ld` and `mull-xctest` are separated processes, so `mull-xctest` doesn't have bitcode information at all. So I want to remove dependencies of llvm::Module related types in Reporter and Executor.

AlexDenisov:
Hi @kateinoigakukun, thanks for sharing the details. I'll get back to you with an elaborate answer later this week.
I definitely see your point and I believe it is related to the #778, so I think there is much more overlap between your work and what we currently have implemented.

kateinoigakukun:
@AlexDenisov Thanks for your pointer!  #778 seems strongly related to my works 👀 

And while I was working on my experimental project, I found another concern about supporting multiple languages. 
mull filters out mutants based on AST information, but this feature depends on language-specific library like libClang.

In Swift case, I'm using [swift-syntax](https://github.com/apple/swift-syntax)  to get AST info, but it's written in Swift, so the mull binary needs to depend on Swift runtime library. Is this dependency acceptable for mull project?

AlexDenisov:
Hi @kateinoigakukun, I appreciate your patience :)

Here are some comments split by topics.

1. Refactoring
I agree we should cut LLVM related things from the reporters (and also from the `mull-xctest`, or `mull-runner`, to be more general). Though, I'd love to preserve more information, not only source location.
We have `MutationPoint` and `Mutant` representing a mutation on the LLVM level and on the source code level, respectively. Currently, `Mutant` is bound to the `MutationPoint`, mostly for historical reasons (i.e. the code is written this way).
With that being said, I suggest we find a way to separate `Mutant` from `MutationPoint`, which is doable but requires some changes across the whole codebase.
Then, we can recreate the `Mutant` from the mutation identifier (`<mutator>:<source location>`).

2. `mull-ld` etc.
Currently, `mull-cxx` only accepts an executable with an embedded bitcode (again, for historical reasons).
IMO, it gives the best user experience... if it works, which is not always the case. There is the same problem on macOS with C++ projects: most of the precompiled dependencies do not have embedded bitcode. So you are not the only one who faces this problem :)
To simplify this part of the process we could either modify `mull-cxx` or create another program that would accept a bitcode file, apply mutations, and emit an object file. I guess this is what your `mull-ld` does?

3. Swift.
First of all, we can definitely add a dependency on Swift. Though, I'd make it optional and put it under some switch (e.g. `cmake -DMULL_ENABLE_SWIFT` or something like this).
I don't yet know how you integrate SwiftSyntax, so I'm curious to learn more about it.
And the last Swift-related point - I guess you want a slightly different set of mutation operators?

---

If you want to have more real-time discussion, please be our guest at [Discord](https://discord.gg/Hphp7dW) server (come over to the `#mull` channel) :)

kateinoigakukun:
> 1. Refactoring
>    With that being said, I suggest we find a way to separate `Mutant` from `MutationPoint`, which is doable but requires some changes across the whole codebase.
>    Then, we can recreate the `Mutant` from the mutation identifier (`<mutator>:<source location>`).

It makes more sense 👍 I'll try to separate them in a few weeks.

> 2. `mull-ld` etc.
>    Currently, `mull-cxx` only accepts an executable with an embedded bitcode (again, for historical reasons).
>    IMO, it gives the best user experience... if it works, which is not always the case. There is the same problem on macOS with C++ projects: most of the precompiled dependencies do not have embedded bitcode. So you are not the only one who faces this problem :)

Oh, I didn't know the issues related to precompiled modules. I'm glad to know I'm not the only one 😂

>    To simplify this part of the process we could either modify `mull-cxx` or create another program that would accept a bitcode file, apply mutations, and emit an object file. I guess this is what your `mull-ld` does?

My `mull-ld` is designed to pretend to be a linker in build system. (for example, `mull-ld` can be passed to CMAKE_LINKER) So it accepts any linker inputs in arguments. Then it filters object files from linker arguments, extracts embedded bitcode section from object files, applies mutations, emits object files for each bitcode module, and launches the actual linker with compiled object files and given linker options.

This pipeline is well similar to mull-cxx, but input and output file types are completely different, so I think it's better to separate these tools.


> 3. Swift.
>    First of all, we can definitely add a dependency on Swift. Though, I'd make it optional and put it under some switch (e.g. `cmake -DMULL_ENABLE_SWIFT` or something like this).
>    I don't yet know how you integrate SwiftSyntax, so I'm curious to learn more about it.


That's good news! Yeah, making it optional makes sense. Unfortunately, Swift support in CMake has some issues, so I had to work around and use hacks 😢 ([e.g. needs to write custom_command to produce object files from Swift code](https://github.com/kateinoigakukun/mull-xctest/blob/6f2da36efcc9c523dbef42c36120cbc48a2d2993/lib/SwiftSupport/core/CMakeLists.txt#L8-L38))




>    And the last Swift-related point - I guess you want a slightly different set of mutation operators?

Currently, [I'm reusing some mutators for C++ and also using Swift specific ones](https://github.com/kateinoigakukun/mull-xctest/blob/6f2da36efcc9c523dbef42c36120cbc48a2d2993/tools/mull-ld/mull-ld.cpp#L262-L265). But I concluded that Swift's LLVM IR generation is very different from Clang. For example, Swift compiler emits `a <= b` as `!(a > b)` and it's expressed using `icmp sgt a, b` in LLVM IR level.  So I think we need different mutators for Swift.

Issue #798, opened 2020-12-23
PSA: Moving away from JIT


AlexDenisov:
Just wanted to share the plans for the nearest future: after lots of discussions and considerations, we decided to move away from JIT to a more classical model.
Just a few words on how everything works right now: Mull extract LLVM Bitcode from an executable, mutates some  LLVM instructions, compiles the bitcode into machine code, and feeds the machine code to a JIT engine for further execution.
This approach works pretty well, sometimes. But often it just fails for some odd reasons.
During the years JIT was the biggest driving force behind the development, but it was also the major source of frustration:
 - simple cases work very well, while the real-world applications don't
 - LLVM's JIT engine is the biggest moving target: we had to put a lot of effort to maintain the same API across different LLVM versions
 - constraints imposed by the JIT engine also limit Mull's applicability: we cannot support as many OSes and language runtimes as we want (think of Windows and Objective-C)
 - debugging is nearly impossible: if something crashes inside of JITted code there is no way to attach a debugger or obtain a reasonable stack trace.
 - etc etc etc...

I'm only slightly exaggerating: the issues I mentioned can be solved, but it becomes a question of trade-offs. In the end, we are not building a JIT engine, we are building a tool for mutation testing.

Short summary of the pros and cons of moving away from JIT.
What we lose:
 - explicit test frameworks support: Mull would still support any test framework, but the granularity and "accuracy" will be lower
 - performance: it needs to be measured first, but I expect slow-down of mutated code under certain conditions
 - custom code coverage: current approach allows us to mutate less code and therefore decrease execution time

What we gain:
 - multithreaded app support
 - support for integration tests: I can certainly relate to what @ligurio asked for in #778
 - better support for various platforms: think of Windows, Linux distros besides Ubuntu, maybe we can get FreeBSD back?
 - better debugging: you get a normal, standalone executable
 - potential to support more languages besides C and C++
 - 32bit apps support: there is still an area where 32bit apps are in use
 - (_my favorite part_) remove lots of half-baked semi-working code I wrote
 - (_another favorite of mine_) we can close lots of JIT related issues (and perhaps open many new ones 😄)

---

The intention so far is not to rewrite Mull from scratch, but gradually migrate to a new model while preserving APIs where possible.

There is no ETA, and there is no clear and detailed plan at the moment. I do have a rough idea of what needs to be done, but it is too rough to be shared in a productive way.

Do not hesitate to drop any questions or thoughts below.

AlexDenisov:
[...]

PR #864, opened 2021-05-13
AST-level mutations using mull-cxx-frontend Clang plugin

stanislaw:
Now that the legacy of LLVM JIT is gone (see #798) and there is a separate `mull-runner` program, it seems doable to finally implement the mutations on the AST level.

This PR introduces a `mull-cxx-frontend` CMake target that creates a Clang plugin that finds mutations in the translation units of a user program as it gets through a normal compilation process (C/C++ languages). The resulting program is a `hypermutant`, which is a single user program executable with all the mutations compiled in it and hidden under the flags that are controlled by passing the environmental variables. The role of the Clang plugin is to find the mutation points and mutate the AST to create and inject the mutations into the executable. The compiled executable is then run by a separate `mull-runner` program - an already existing solution that have been implemented recently.

Credits: this is an implementation of the `hypermutant` idea that Alex Denisov has found in one of the papers (TBD: check which one). Without that paper, we might not have invented such a solution ourselves or at least not have started it now.

The Clang API is not very friendly when it comes to mutating the AST tree, but it is still possible with an addition of several workarounds.

---

The following tasks are still pending:

- [x] Implement all mutations from https://mull.readthedocs.io/en/latest/HowMullWorks.html#supported-mutation-operators.
- [x] Make sure the solution can run on macOS and Linux for all LLVM versions. The focus point: AST mutation-related calls to Clang API.

The following tasks might be done in the separate PRs:

- [ ] Make Mull Runner read the AST information from the binary, not from the compilation database (new source code provider).
- [ ] Support incremental testing.
- [ ] Simplify a distribution of a plugin by including it to the binary distributions of Mull.
- [ ] For CMake: implement necessary boilerplate to support finding Mull's plugin and runner via `find_package(Mull)...`.
- [ ] Passing the command-line arguments to the Clang plugin. 
  - The Clang plugin args system is quite verbose, so it could be reasonable to move to a YAML config file-based approach.
  - It would be great to introduce a conventional mapping with the AOR-like naming scheme.


tools/mull-cxx-frontend/src/MullClangPlugin.cpp
```diff
@@ -0,0 +1,132 @@
...
+namespace {
+
+class MullASTConsumer : public ASTConsumer {
+  CompilerInstance &Instance;
+  std::unique_ptr<MullASTMutator> astMutator;
+  MutationMap mutationMap;
+
+public:
+  MullASTConsumer(CompilerInstance &Instance, const MutationMap mutationMap)
+      : Instance(Instance), astMutator(nullptr), mutationMap(mutationMap) {}
```
AlexDenisov, line 27: Nitpick: we don't use Uppercase names.

tools/mull-cxx-frontend/src/ASTMutationsSearchVisitor.cpp
```diff
@@ -0,0 +1,215 @@
...
+    return;
+  }
+  std::string sourceFilePath = sourceManager.getFilename(location).str();
+  if (sourceFilePath.empty()) {
+    /// we reach here because of asserts()
+    /// TODO: maybe there are more cases.
+    /// assert(0);
+    return;
+  }
+  if (sourceFilePath.find("include/gtest") != std::string::npos) {
```
AlexDenisov, line 197: I guess some copy-paste? :)

tools/mull-cxx-frontend/src/MullClangPlugin.cpp
```diff
@@ -0,0 +1,132 @@
...
+      astMutator->instrumentTranslationUnit();
+    }
+
+    for (DeclGroupRef::iterator I = DG.begin(), E = DG.end(); I != E; ++I) {
+      if ((*I)->getKind() != Decl::Function) {
+        continue;
+      }
+
+      FunctionDecl *f = static_cast<FunctionDecl *>(*I);
+      if (f->getDeclName().getAsString() == "main") {
```
AlexDenisov, line 56: Let's mutate everything and find another way to filter out certain functions?

tools/mull-cxx-frontend/src/MullClangPlugin.cpp
```diff
@@ -0,0 +1,132 @@
...
+        continue;
+      }
+
+      clang::SourceLocation functionLocation = f->getLocation();
+      if (Instance.getSourceManager().isInSystemHeader(functionLocation)) {
+        continue;
+      }
+      std::string sourceFilePath = Instance.getSourceManager().getFilename(functionLocation).str();
+      if (sourceFilePath.find("include/gtest") != std::string::npos) {
```
AlexDenisov, line 65: Do we need this immediately? i.e. can we mutate everything and then add filters at a later stage?

AlexDenisov (approved):
There are a few nitpicks. Feel free to ignore them, but I must say they distract from the actual code review :)

All in all, let's merge it as soon as you think it is ready for the merge, and then work on the improvements :)

.gitignore
```diff
@@ -18,4 +18,5 @@ cmake-build*
 *.retry
 tests-lit/tests/**/*.ll
 
+.DS_Store
```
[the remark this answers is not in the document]
stanislaw: As discussed in person, we better gitignore this because it still shows up in Docker containers and Vagrant machines.

tools/mull-cxx-frontend/src/ClangASTMutator.cpp
```diff
@@ -0,0 +1,153 @@
...
+      /// things play against current Clang AST API.
+      /// TODO: Find a better way to perform the mutation.
+      clang::Stmt *parentStmt = (clang::Stmt *)constParentStmt;
+      clang::Stmt::child_iterator parentChildrenIterator =
+          std::find(parentStmt->child_begin(), parentStmt->child_end(), oldExpr);
+      assert(parentChildrenIterator != parentStmt->child_end());
+      *parentChildrenIterator = conditionalExpr;
+      return;
+    } else if (const clang::VarDecl *constVarDecl = p.get<clang::VarDecl>()) {
+      clang::VarDecl *parentVarDecl = (clang::VarDecl *)constVarDecl;
```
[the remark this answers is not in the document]
stanislaw: 😸

tools/mull-cxx-frontend/src/ASTMutationsSearchVisitor.cpp
stanislaw: Certainly not. It helps a lot during testing against a bigger codebase. For any GoogleTest-based project, it is very useful for now. I would really prefer to keep this until we have the filters in place. 

The same comment about the `main()` function below.

tools/mull-cxx-frontend/src/MullClangPlugin.cpp
stanislaw: The same answer as for the previous comment - let's keep it until we have filters!

One of these filters filters gtest's translation units, the other one the inlined code, so they are both useful for now.

lib/AST/ASTConstants.cpp
```diff
@@ -0,0 +1,49 @@
+#include "mull/AST/ASTConstants.h"
+
+std::vector<
+    std::tuple<clang::BinaryOperator::Opcode, mull::MutatorKind, clang::BinaryOperator::Opcode>>
+    mull::BINARY_MUTATIONS = {
```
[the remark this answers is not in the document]
stanislaw: No-brainer, changed.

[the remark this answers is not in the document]
stanislaw: As discussed in person, I will take care of this later. We need to understand when a global variable creates problems with the order of initialization.

tools/mull-cxx-frontend/src/ASTMutationsSearchVisitor.cpp
```diff
@@ -0,0 +1,215 @@
+#include "ASTMutationsSearchVisitor.h"
+
+#include "MutationMap.h"
+
+#include "mull/AST/ASTConstants.h"
+
+#include "mull/AST/MullClangCompatibility.h"
+
```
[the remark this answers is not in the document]
stanislaw: No-brainer, changed.

tools/mull-cxx-frontend/src/MullClangPlugin.cpp
stanislaw: No-brainer, changed.

tools/mull-cxx-frontend/src/MullClangPlugin.cpp
stanislaw: See previous comment.

PR #884, opened 2021-06-13
Decrease the binary size

AlexDenisov:


AlexDenisov:
Drops the package size by ~3.5-4Mb.

PR #901, opened 2021-10-03
Fix tests on macOS and lowercase most of the lit substitutions

AlexDenisov:


AlexDenisov:
Some tests are still broken when run locally against macOS 11+ SDK, but I'm merging this as it at least improves the naming convention

PR #910, opened 2021-10-31
Remove last reference to test-framework

m42e:
There was one reference to the `-test-framework` command line option in the documentation.


AlexDenisov:
Thank you, @m42e!

Issue #775, opened 2020-10-16
On the way to v1.0


AlexDenisov:
This issue is supposed to be a place to discuss what needs to be implemented for version 1.0.
These should include show-stoppers but can also include nice-to-haves. We'll need to decide which ones belong to which group.

Everyone is encouraged to share their ideas and suggestions. I'll start.

---

 - **Multithreading support**: mull itself is multithreaded, but if the program under test is multithreaded, then Mull may produce non-deterministic results. Currently, Mull gathers the code coverage based on call trees. In a single-threaded program (most of the time) the call tree is deterministic (i.e., the order of the calls is the same from run to run). A possible solution is to use another, 'linear' coverage approach: to only collect whether the function was executed or not. It has some drawbacks but will give the most deterministic behavior.
 - **TLS support**: this is being in the works currently #745.
 - **Diff reporter**: Mull has several reporters (mutation elements, SQLite, CLI). it would be very nice to also have a reporter that produces git-like patches so that they can be applied to the source code or used as part of CI integration. Related #587.
 - **CI integration**: mull can be run as part of the CI process, but the integration is not trivial and does not have smooth CI integration.  Directly related to #738.
 - **Diff-based mutation filter**: at the moment, Mull does not respect version control systems. For better CI integration it is essential to filter out mutants that are unrelated to a pull or merge request.
 - **Non covered mutants**: as described here #636.
 - **More test frameworks**: it would be nice to have more test frameworks supported. Related #174 #175 #191 #492 #746. More suggestions are very welcome.

---

The above list is not exhaustive and may be extended.

AlexDenisov:
- **Smooth junk detection**: currently, Mull requires a user to provide 'implicit' C/C++ header search paths manually. It would be nice if we can automate this process in a transparent way.

gebart:
Great to collect all the important features in one ticket. Would it be useful to also create a v1.0 Label or Milestone to mark PRs and issues which are part of this effort?

AlexDenisov:
That's a good idea. I'll create a milestone and a label. I guess for the milestone to work I'd also need to create an issue per each feature/task.

chgans:
About the diff filter, are we talking something like clang-tidy-diff?
https://clang.llvm.org/extra/doxygen/clang-tidy-diff_8py_source.html

AlexDenisov:
> About the diff filter, are we talking something like clang-tidy-diff?
[...]

Issue #587, opened 2019-10-16
is it possible to save the source code level mutant record? I want to analysis them #10198


Marvinmw:
is is possible to save the source code level mutant record? I want to analysis them #10198

AlexDenisov:
Unfortunately not at the moment. There is something in the works and on our todo list, though.

I'll notify you when we have this in place.

ligurio:
Mutations in source code are part of Mull report (type Elements). It is not so difficult to create real patches from these changes.
(btw I do it using https://github.com/ligurio/py-mutation-testing-elements)

AlexDenisov:
It took some time to get there, thanks @m42e for the effort!
@Marvinmw this is resolved now (via #911).

PR #911, opened 2021-10-31
Patch reporter

m42e:
Solves https://github.com/mull-project/mull/issues/587
Hope this helps to get closet to V1.0 (https://github.com/mull-project/mull/issues/775)

### DIFF
--- a/docs/command-line/generated/mull-cxx-cli-options.rst
+++ b/docs/command-line/generated/mull-cxx-cli-options.rst
@@ -20,6 +20,8 @@
 
     :Elements:	Generates mutation-testing-elements compatible JSON file
 
+    :Patches:	Generates a patchfile for each mutation. uses `--git-project-root` as base dir.
+
 --ide-reporter-show-killed		Makes IDEReporter to also report killed mutations (disabled by default)
 
 --debug		Enables Debug Mode: more logs are printed
--- a/docs/command-line/generated/mull-runner-cli-options.rst
+++ b/docs/command-line/generated/mull-runner-cli-options.rst
@@ -16,6 +16,8 @@
 
     :Elements:	Generates mutation-testing-elements compatible JSON file
 
+    :Patches:	Generates a unified patchfile for each mutation
+
 --ide-reporter-show-killed		Makes IDEReporter to also report killed mutations (disabled by default)
 
 --debug		Enables Debug Mode: more logs are printed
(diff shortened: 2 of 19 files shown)


m42e:
Suggestion: use `--git-project-root` as base directory. What do you think? @stanislaw  @AlexDenisov

AlexDenisov:
Thank you so much for bringing this up! This is certainly a great start!

Could you add some test cases for some mutation operators? In particular, the interesting operators are `cxx_lshift_to_rshift`, `cxx_lt_to_ge`, `cxx_le_to_lt` (at least).

The other important one is `cxx_remove_void_call`, but I don't think it's possible to get it right without actually parsing the source code.

> Suggestion: use --git-project-root as base directory. What do you think?

What is the base directory in this context?

m42e:
Ok, will do i can add some sample files, too. 

Base path means the Path the patch should be relative to. For now it is absolute.

m42e:
The test should use simply Google test or is there another way these check should be done?

AlexDenisov:
There are integration tests based on LLVM lit and filecheck under `tests-lit` directory, I'd recommend adding the tests there.

AlexDenisov:
Oh, and the git root sounds good as a base path!

m42e:
> Oh, and the git root sounds good as a base path!

I decided to use the git path as fallback for `mull-cxx` and added an extra command line parameter, as `mull-runner` does not have the `git-project-dir` option. Feel free to suggest otherwise. But I think it might be a bit misleading in the `mull-runner` context if the `git-project-dir` is not used otherwise.

AlexDenisov:
Just played around with it locally and it looks pretty good!
We'll need to mention somewhere that it only works for single-line changes (i.e. multiline patches won't work correctly) and that some mutators may produce invalid code (at least `cxx_init_const` and `negate_mutator` don't work properly).

All in all, looks good and I'm happy to merge unless you want to add something :)

m42e:
Feel free to merge. Also feel free to send me examples of what’s not working an I will have a look at it.

AlexDenisov:
> Also feel free to send me examples of what’s not working an I will have a look at it.

`cxx_const_init` doesn't work properly with [the hello-world example](https://mull.readthedocs.io/en/latest/tutorials/HelloWorld.html).
`remove_void_function` produces incorrect diff for something like
```c
void foo(...);
///
foo(a,
      b);
```
and I guess
```c
void foo();
void bar();
///
foo(); bar();
```
will also be incorrect.

All in all, there is no straightforward way to handle this at the moment: we need to attach certain metadata to some mutation points.

m42e:
@AlexDenisov I tried the `cxx_init_const` and I think it is actually the replacement information or the source location information that is not correct. It is also not displayed correctly using the `Elements` reporter.

The multiline does not work, that's sadly true, fix is here: #913

the third example works just fine.

PR #915, opened 2021-11-12
Fixed order of drain and wait, as proposed in the example of reproc

m42e:
The incorrect order of drain and wait causes an incorrect recognition of failed original test and prolonged timeouts.

See https://github.com/DaanDeMeyer/reproc/blob/main/reproc%2B%2B/examples/drain.cpp#L50


m42e:
@AlexDenisov Looks like the download failed, do you have an idea why that happens?

AlexDenisov:
> @AlexDenisov Looks like the download failed, do you have an idea why that happens?

Sadly, it happens once in a while. I restarted the job.

There is another timeout-related test failing It would be nice if we can fix it.

Issue #845, opened 2021-03-23
Mull crashed with incorrect regex passed to --include-path


ligurio:
### How-to reproduce

```
sergeyb@pony:~/sources/mull_testing$ mull-cxx --include-path=*func1.c ./main
[info] Extracting bitcode from executable (threads: 1)
       [################################] 1/1. Finished in 1ms
[info] Loading bitcode files (threads: 3)
       [################################] 3/3. Finished in 12msterminate called after throwing an instance of 'std::regex_error'
  what():  regex_error
Aborted (core dumped)
```

With gdb:
```
(gdb) run --include-path=*func1.c ./main
Starting program: /usr/bin/mull-cxx --include-path=*func1.c ./main
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[info] Extracting bitcode from executable (threads: 1)
       [################################] 1/1. Finished in 1ms[New Thread 0x7ffff5a2b700 (LWP 2445803)]
[New Thread 0x7fffeffff700 (LWP 2445804)]
[New Thread 0x7ffff522a700 (LWP 2445805)]

[info] Loading bitcode files (threads: 3)
[New Thread 0x7ffff4a29700 (LWP 2445806)]
       [#####################-----------] 2/3[Thread 0x7fffeffff700 (LWP 2445804) exited]
[Thread 0x7ffff5a2b700 (LWP 2445803) exited]
[Thread 0x7ffff522a700 (LWP 2445805) exited]
       [################################] 3/3. Finished in 11ms[Thread 0x7ffff4a29700 (LWP 2445806) exited]
terminate called after throwing an instance of 'std::regex_error'
  what():  regex_error

Thread 1 "mull-cxx" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb)
```

### Version

Mull: LLVM-based mutation testing
https://github.com/mull-project/mull
Version: 0.10.0-trunk1
Commit: eed5960
Date: 22 Mar 2021
LLVM: 11.0.0

PR #918, opened 2021-11-13
detect invalid regex

m42e:
As this is only possible by catching the exception the flag
-fno-exception has been removed

Fixes: https://github.com/mull-project/mull/issues/845

### DIFF
--- a/include/mull/Filters/FilePathFilter.h
+++ b/include/mull/Filters/FilePathFilter.h
@@ -28,8 +28,14 @@ public:
 private:
   bool shouldSkip(const mull::SourceLocation &location) const;
 
+// The API of llvm::Regex support const match starting from 10.0.0
+#if LLVM_VERSION_MAJOR >= 10
+  std::vector<llvm::Regex> includeFilters;
+  std::vector<llvm::Regex> excludeFilters;
+#else
   mutable std::vector<llvm::Regex> includeFilters;
   mutable std::vector<llvm::Regex> excludeFilters;
+#endif
 
   mutable std::unordered_map<std::string, bool> cache;
   mutable std::mutex cacheMutex;
--- a/include/mull/Filters/FilePathFilter.h
+++ b/include/mull/Filters/FilePathFilter.h
@@ -4,17 +4,17 @@
 #include "mull/Filters/InstructionFilter.h"
 #include "mull/Filters/MutationFilter.h"
 
+#include <llvm/Support/Regex.h>
 #include <mutex>
-#include <regex>
+#include <string>
+#include <utility>
 #include <unordered_map>
 #include <vector>
 
 namespace mull {
 struct SourceLocation;
 
-class FilePathFilter : public MutationFilter,
-                       public FunctionFilter,
-                       public InstructionFilter {
+class FilePathFilter : public MutationFilter, public FunctionFilter, public InstructionFilter {
 public:
   bool shouldSkip(MutationPoint *point) override;
   bool shouldSkip(llvm::Function *function) override;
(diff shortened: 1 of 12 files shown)


stanislaw (changes requested):
I haven't aligned my view on this changeset with Alex, but I think he would agree with the following:

1) Mull does not use exceptions in its current design. Could you implement the same functionality without using exceptions?

2) Please provide a LIT/FileCheck test that exercises this implementation path.

m42e:
1. there is neither a non throwing way within the standard library, not could I think of constructing a regex to check a regex completely. There sure is but requires back references for sure. So the answer is: no.
2. can do, but check first if you want to stay with the "no exceptions" rule.

m42e:
You could still break the Interface and accept globbing patterns inständig of regex.

stanislaw:
I would be curious to know if LLVM has its own regex library. We have been using several things from LLVM sys library instead of the C++ std. Any chance to find something there?

m42e:
Ok, if using llvm regex is ok for you we could switch to that. From a first glance that looks promising.

m42e:
Finals, regex check, with error description and without exceptions.

lib/Filters/FilePathFilter.cpp
```diff
@@ -55,15 +55,30 @@ bool FilePathFilter::shouldSkip(const std::string &sourceFilePath) const {
...
+std::pair<bool, std::string> FilePathFilter::exclude(const std::string &filter) {
 
-void FilePathFilter::exclude(const std::string &filter) {
-  excludeFilters.emplace_back(filter, std::regex::egrep);
+  auto regex_filter = llvm::Regex(filter);
+  std::string error;
+  if (!regex_filter.isValid(error)) {
+    return std::make_pair(false, std::move(error));
+  }
+  excludeFilters.emplace_back(filter);
```
AlexDenisov, line 72: Shouldn't it contain `regex_filter`?

include/mull/Filters/FilePathFilter.h
```diff
@@ -4,32 +4,32 @@
...
-  void include(const std::string &filter);
+  std::pair<bool, std::string> exclude(const std::string &filter);
+  std::pair<bool, std::string> include(const std::string &filter);
 
 private:
   bool shouldSkip(const mull::SourceLocation &location) const;
 
-  std::vector<std::regex> includeFilters;
-  std::vector<std::regex> excludeFilters;
+  mutable std::vector<llvm::Regex> includeFilters;
```
AlexDenisov, line 31: Why does it have to be mutable?

AlexDenisov (commented):
This is fantastic work! There are two minor questions and we are good to go!

include/mull/Filters/FilePathFilter.h
m42e: Yes, it does. llvm::Regex does not offer const qualified member functions.

lib/Filters/FilePathFilter.cpp
m42e: You are right, I messed this up, because I was trying to push back the string and create the regex each time it will get checked. But making it mutable seems to make more sense here.

include/mull/Filters/FilePathFilter.h
AlexDenisov: Do you know from which versions it provides the const qualified functions? If it's only some "old" versions, can we wrap it in `#if/#endif` and drop later when we drop those versions support?

lib/Filters/FilePathFilter.cpp
AlexDenisov: Based on the CI failure, I think these should be `std::move`d.

include/mull/Filters/FilePathFilter.h
m42e: If you prefers it that way I can have a look

AlexDenisov: That would be really nice! I had some hard time debugging multithreading issues with `mutable` objects hidden down the line, so I'd prefer to avoid them whenever possible 😄

m42e: the mutable seems to only be necessary as the interface itself is defined as const.

lib/Filters/FilePathFilter.cpp
m42e: If I remember correctly move ctor has been deleted. I will have a look.

AlexDenisov: Ok, if that's the case, then we can just construct each regex twice, should be fine.

m42e:
It compiled with 6.0 and 13.0 and seems to work.

AlexDenisov (approved):

PR #919, opened 2021-11-17
minor fix for path in patches

m42e:
Path was not quite correct, missing a starting slash


m42e:
will merge after the multiline path is merged.

PR #923, opened 2021-11-22
Enable more detailed coverage

m42e:
This enables branch based coverage elimination.
Please check if this matches your expectation. I would suggest to have an additional switch to enable the line based coverage as it is quite a break in the behaviour compared to the current one.

Let me know what you think.

### DIFF
--- a/include/mull/FunctionUnderTest.h
+++ b/include/mull/FunctionUnderTest.h
@@ -1,6 +1,8 @@
 #pragma once
 
+#include "mull/SourceLocation.h"
 #include <vector>
+#include <llvm/ProfileData/Coverage/CoverageMapping.h>
 
 namespace llvm {
 class Function;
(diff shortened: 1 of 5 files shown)


AlexDenisov:
It's ok to break the current behavior, but I'd strongly recommend adding/extending the suite to reflect the changes.

m42e:
For sure, but just wanted some ok before doing so.

m42e:
I had something prepared (manual testing so far) but should confirm it is working.

m42e:
while scrolling through the code, would you consider something like this more clear @AlexDenisov:

```c++
bool FunctionUnderTest::isCovered(const SourceLocation& location) const {
  if (linecoverage.size() == 0){
    return covered;
  }
  bool iscovered = true;
  auto after_start = [&location](auto LineStart, auto ColumnStart){
    return (LineStart < location.line || (LineStart == location.line && ColumnStart <= location.column));
  };
  auto before_end = [&location](auto LineEnd, auto ColumnEnd){
    return (LineEnd > location.line || (LineEnd == location.line && ColumnEnd >= location.column));
  };
  for_each(linecoverage.begin(), linecoverage.end(), [&](const auto & cov){
    if (after_start(cov.LineStart, cov.ColumnStart) && before_end(cov.LineEnd, cov.ColumnEnd)){
      iscovered &= cov.ExecutionCount > 0;
    }
  });
  return iscovered;
}

```

PR #1000, opened 2022-05-05
Add slow verification for debugging

AlexDenisov:
Mull verifies that the module is valid before applying any mutations, and after applying all the mutations.
it is now possible to add verification after each mutation to find the exact offensive mutation operator.
Verifying the module after each mutation adds significant overhead, so this option is disabled by default and can be enabled via the config file:
```yml
debug:
  slowIRVerification: true
```


ligurio:
It is worth to describe the option in documentation.

PR #1012, opened 2022-11-06
Address misc. build issues and warnings

vengaer:
While experimenting with mull, I ran in to a few issues when building. The only "serious" hurdle was a missing header causing compilation to fail. While at it, I took the liberty of fixing a few warnings as well.

I figured I'd share my changes in the event that you would like to incorporate them upstream.

### DIFF
--- a/include/mull/Filters/GitDiffReader.h
+++ b/include/mull/Filters/GitDiffReader.h
@@ -9,7 +9,7 @@
 namespace mull {
 class Diagnostics;
 
-typedef std::pair<int, int> GitDiffSourceFileRange;
+typedef std::pair<unsigned, unsigned> GitDiffSourceFileRange;
 typedef std::vector<GitDiffSourceFileRange> GitDiffSourceFileRanges;
 typedef std::map<std::string, GitDiffSourceFileRanges> GitDiffInfo;
 
--- a/lib/Config/ConfigurationOptions.cpp
+++ b/lib/Config/ConfigurationOptions.cpp
@@ -1,5 +1,6 @@
 #include "mull/Config/ConfigurationOptions.h"
 
+#include <algorithm>
 #include <thread>
 
 namespace mull {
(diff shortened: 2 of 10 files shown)


AlexDenisov:
Hi @vengaer, thank you so much for sending this PR, highly appreciated!

Unfortunately, some of the changes break under different LLVM versions.
Although, I think the macOS failure is not related to your change, I'll take a look into it.

vengaer:
There, that should deal with the warning as intended while remaining compatible with older llvm versions. 

I don't think there's anything else I've broken but I'll keep an eye on the workflows in case there is. Let me know if you have any opinions on the changes themselves.

AlexDenisov:
Yeah, the last commit looks a bit more hairy than I'm comfortable with. I think in this case "simple" `#if` guard would be a better and more pragmatic choice.

vengaer:
Sure thing. Didn't even occur to me that LLVM would define version macros but it makes perfect sense now that you mention it. 

I take that to mean that compatilibty with "only" official LLVM releases is enough. As in there being no need to support arbitrary commits between, say, LLVM 13 and 14. Feel free to correct me if I'm mistaken.

vengaer:
ICYMI I've changed the last commit to use an ifdef as requested. It should pass the Linux builds. The Mac issues last time seem to have been caused by Ansible not being available.

I don't mean to rush you, it's entirely up to you if and when you want this merged. I'm perfectly fine with using my fork for the time being. Just don't want it to seem like I've lost interest in seeing this through.

AlexDenisov:
Hi @vengaer, I appreciate your patience 🙇

The macOS fix is here https://github.com/mull-project/mull/pull/1013/files, you can rebase/merge your branch with and we should be good to go.

vengaer:
Great, thanks for fixing it! Should be done, fingers crossed!

PR #1016, opened 2023-01-14
Fall back on clang-cpp.so if libclangTooling.so is unavailable

vengaer:
I ran into some linking issues when building Mull on Arch due to libclangTooling.so being unavaiable. After doing some digging I came upon [this bug report](https://bugs.archlinux.org/task/66283) related to Arch's packaging of clang. Turns out its not a bug at all and that they are simply building clang without `BUILD_SHARED_LIBS=on`. When this flag is not passed, the symbols end up in libclang-cpp.so instead. 

This change makes the build system handle both options by first attempting to find libclangTooling.so and, if it's missing, fall back on libclang-cpp.so instead.

### DIFF
--- a/cmake/functions.cmake
+++ b/cmake/functions.cmake
@@ -28,3 +28,49 @@ macro(add_mull_executable)
     RUNTIME DESTINATION bin
   )
 endmacro()
+
+function(determine_clang_component_lib lib_out component)
+  # Assume the component comes as a separate lib
+  # until otherwise proven
+  set(${lib_out} ${component} PARENT_SCOPE)
+
+  # Check whether the LLVM shared library was built
+  if("${LLVM_LINK_LLVM_DYLIB}" STREQUAL "ON" AND NOT "${LLVM_DYLIB_COMPONENTS}" STREQUAL "")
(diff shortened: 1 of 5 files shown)


AlexDenisov:
Hi @vengaer, thanks for the fix.
I'm in favour in accepting this PR as long as we can make it work on Ubuntu as well. I can tweak macos parts when the linux side is done.

AlexDenisov:
As a side note, we can probably add Arch Linux to the CI if there is enough interest.

vengaer:
Working on it! Sorry for the noise. Seems I've misunderstood the search paths used by `target_link_libraries` and `find_library`.

I'm open to adding Arch if you want it, although that would have to be it's own PR. I'm not at all familiar with vagrant so it'd require a bit of reading up on my part.

vengaer:
Seems I was quite a bit off in my understanding regarding what decides which libs are built when. Can't say I have the full picture even now but I'm quite confident this should prove pretty robust.

I ended up adding an Ubuntu 20.04 VM for testing. While at it, I removed the Ubuntu 16 VM since it doesn't support any of the LLVM versions Mull is targeting

PR #1017, opened 2023-01-14
Disallow surviving mutants in strict mode

vengaer:
The --help for strict mode states that "all warning messages are treated as fatal errors". While this is indeed true for anything that goes through Diagnostics::warning, there are a few messages presented as warnings that are not written via said function. The one (and at least so far only) concrete example I have come across are warnings written in printMutant in IDEReporter.cpp which emits messages similar to

```text
<file>:144:17: warning: Survived: Replaced > with >= [cxx_gt_to_ge]
    if(currsize > size) {
                ^
```

Given the description for the --strict flag, it would seem as though the above should be considered a fatal error too. 

The warnings written in printMutant are currently written through fprintf. While it would be simple to change this this to use Diagnostics::warning instead, that doesn't seem to be the best approach. Seeing as all the mutants have already run, all that would achieve is cut parts of the report.

Instead of omitting useful information, this checks the number of surviving mutants at the very end of mull-runner's main. If any survived, this is logged this via Diagnostics::warning. This way, mull-runner will exit with a non-zero status if there are surviving mutants in strict mode. If not in strict mode, mull-runner always exits with status 0.

This change would, of course, alter the behavior of the --strict flag which I could see being a deal breaker. If that is the case, perhaps it'd be better to add another command line argument to achieve the intended results. Either way, I'd personally find it very useful to have a way of forcing Mull to exit with an error when there are surviving mutants.

Please let me know if you'd prefer this to be handled differently

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ tests-lit/tests/**/*.ll
 
 .DS_Store
 
+tags
--- a/docs/command-line/generated/mull-runner-cli-options.rst
+++ b/docs/command-line/generated/mull-runner-cli-options.rst
@@ -28,6 +28,8 @@
 
 --strict		Enables Strict Mode: all warning messages are treated as fatal errors
 
+--allow-surviving	Do not treat mutants surviving as an error
+
 --no-test-output		Does not capture output from test runs
 
 --no-mutant-output		Does not capture output from mutant runs
--- a/infrastructure/Vagrantfile
+++ b/infrastructure/Vagrantfile
@@ -3,6 +3,8 @@
 # to upgrade instead of disabling the requirement below.
 Vagrant.require_version ">= 1.7.0"
 
+mull_version = %x(git describe --tags --dirty | tr -d '\n')
+
 def configure_vm(name, config)
   config.vm.provider "virtualbox" do |v|
     v.memory = 8192
@@ -25,6 +27,7 @@ Vagrant.configure(2) do |config|
       ansible.extra_vars = {
         llvm_version: ENV['LLVM_VERSION'] || '6.0.0',
         gitref: ENV['GITREF'] || 'main',
+        mull_version: ENV['mull_version'] || "#{mull_version}"
       }
     end
   end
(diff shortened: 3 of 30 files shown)


AlexDenisov:
The warnings from `IDEReporter` are not the same as warnings coming from Mull itself, so it is expected that they are not affected by the strict mode.

In general, the idea of failing in case of surviving mutants makes sense and is a great addition.
Though, I'd perhaps add another CLI arg to control this behavior, and maybe even enable it by default?
I don't have a strong opinion on how the CLI arg should be called, so I'm open for suggestions.

AlexDenisov:
Although, enabling it by default would likely break many integration tests, so I guess it's better to not turn it on by default just yet - unless you also want to update all the affected tests 😄

vengaer:
> The warnings from IDEReporter are not the same as warnings coming from Mull itself, so it is expected that they are not affected by the strict mode.

I see. Then it's just my getting tricked by the wording :)

> Though, I'd perhaps add another CLI arg to control this behavior, and maybe even enable it by default?
I don't have a strong opinion on how the CLI arg should be called, so I'm open for suggestions.

A separate argument just as well. I'll have a look at it. If the number of failing tests is managable, I'll enable it by default. As for the naming, perhaps something along the lines of --allow-surviving (or --disallow-surviving if not enabled by default)? It's a little verbose but at least it should be obvious what it does.

vengaer:
Or, perhaps, we could simply enable the option by default but explicitly disable it for the integration tests. That would save me quite a bit of work I'd imagine

AlexDenisov:
`--allow-surviving` sounds great!

vengaer:
As discussed, mutants surviving is now treated as an error by default but can be ignored by passing the `--allow-surviving` CLI argument. The lit and integration tests have been updated to accomodate the new behavior.

Also made some minor changes to the infrastructure to make the CI workflow succeed when run via vagrant. Tested with the ubuntu20.04 box.

AlexDenisov:
Thank you so much @vengaer!

vengaer:
No worries! You guys have something special here, happy to help in whatever little way I can

PR #1048, opened 2024-05-22
Add ubuntu 24.04 ci

rjaegers:
Add Ubuntu 24.04 with LLVM 14, 15, 16 and 17

Refactored GitHub workflow to use a matrix for Ubuntu ci.


AlexDenisov:
Thank you once again, this is a great improvement!

PR #1052, opened 2024-07-04
Add a mutation score threshold so that we can still pass without kill…

mpownby:
Add a mutation score threshold so that we can still pass without killing every single mutant.

Motivation:
I want to use mull in my CI/CD pipeline to make sure incoming pull requests meet a minimum testing standard.

Currently, mull can be configured to pass (return exit code 0) if a) all mutants are killed or b) all surviving mutants are ignored.  This is accomplished via the "--allow-surviving" command line option.

I'd like a more granular option.  I might like to have (nearly) all mutations enabled, but killing every single mutant may not be worth it to me, especially if they are inside code that isn't critical like formatting log messages or rare catch blocks.  I'd prefer to pass/fail based on a mutation score threshold.

Hence this pull request.

### DIFF
--- a/tools/mull-runner/mull-runner.cpp
+++ b/tools/mull-runner/mull-runner.cpp
@@ -242,6 +242,22 @@ int main(int argc, char **argv) {
       return resPtr->getExecutionResult().status == mull::ExecutionStatus::Passed;
     });
 
+  // Calculate mutation score for later threshold comparison
+    int killedMutantsCount = 0;
+    int totalMutantsCount = 0;
+
+  for (auto &mutationResult : mutationResults) {
+    auto &executionResult = mutationResult->getExecutionResult();
+
+    totalMutantsCount++;
+    if ((executionResult.status != mull::ExecutionStatus::NotCovered) && (executionResult.status != mull::ExecutionStatus::Passed)) {
+      killedMutantsCount++;
+    }
+  }
+
+  auto rawScore = double(killedMutantsCount) / double(totalMutantsCount);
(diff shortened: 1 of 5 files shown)


AlexDenisov (approved):

mpownby:
> Hi @mpownby, thank for the PR, this is a great idea and a great feature!
> 
> You'd need to update some test cases which are failing, and add a new test case covering this specific feature. You can put the test case somewhere here tests-lit/tests, feel free to simply copy an existing test and modify as needed.

Ok, I think I finally got it.  It was a real conundrum figuring out how to run the lit tests locally but I finally figured it out.  I added an integration test that achieves a mutation score of 50% with a desired threshold of 49% so the runner is verified to still return success.

tools/mull-runner/mull-runner.cpp
```diff
@@ -259,7 +275,7 @@ int main(int argc, char **argv) {
     stringstream << "Surviving mutants: " << surviving;
     diagnostics.info(stringstream.str());
 
-    if (!tool::AllowSurvivingEnabled) {
+    if ((!tool::AllowSurvivingEnabled) && (tool::MutationScoreThreshold > score)) {
```
AlexDenisov, line 278: Could you please add/extend a test case to check when "allow surviving" is enabled and the score is exactly 100? I guess this `>` should be `>=`.

AlexDenisov: You're right, apologies for the confusion, it was a long day 😃

PR #1099, opened 2025-05-02
Support for building with LLVM 19

vengaer:
I ran into some trouble when wanting to use Mull with LLVM 19. Can't claim there is too much interesting in it the incompatibilities - mostly a few functions that have been either renamed or removed, and a few headers that now need to be included explicitly.

I won't claim that there changes include everything required for full LLVM 19 compatibility. I have, for example, made no effort to update the CI so this is definitely not on par with the work done when adding LLVM 18 support. Still, it ought to at least be a decent start. If you're interested in it, you're more than welcome to use it.


AlexDenisov:
Thank you for the initiative @vengaer, I'll try to get to it soon-ish 🙇

AlexDenisov (approved):

PR #1183, opened 2026-04-10
Fix deadlock when mull-runner executes a command with large stdout/stderr

ccotter:
When running a large test suite through mull-runner, it became stuck. The root cause was the `child.wait_timeout` never finishing because the stdout or stderr pipe filled up.

The fix I chose here was to use background threads to read stderr, stdout. They will be short lived threads. Another solution would be to use non blocking IO and polling, but that's trickier to get right, and seems like overkill for this specific function.


AlexDenisov:
Thank you @ccotter, I've cut a new release with the fix https://github.com/mull-project/mull/releases/tag/0.33.0

ccotter:
Thank you for the review!

PR #1185, opened 2026-04-22
Add mull-instrument standalone tool

ccotter:
Per discussion on discord to add a new standalone tool.


tools/mull-instrument/mull-instrument.cpp
```diff
@@ -0,0 +1,68 @@
...
+  llvm::LLVMContext context;
+  llvm::SMDiagnostic err;
+
+  auto module = llvm::parseIRFile(InputFile, err, context);
+  if (!module) {
+    err.print(argv[0], llvm::errs());
+    return 1;
+  }
+
+  auto core = init_core_ffi(DiagOutput::Stderr);
```
ccotter, line 51: Originally, I did not lift init_core_ffi out, but the actual reason I lifted it out was `mutateBitcode` wrote to stdout, which polluted `WriteBitcodeToFile` if it was also writing to stdout. Hence the new `DiagOutput` enum.

I also considered changing the diagnostics to go to stderr instead of stderr without an option.

ccotter:
On discord, @AlexDenisov suggested a rust tool instead, though it wasn't clear to me what the benefits were vs FFI gymnastics to talk to the existing LLVM/C++ APIs, as the main job of this new tool is to invoke what is mostly C++ anyway. In the spirit of trying to keep it simple, I left the tool as C++.

AlexDenisov:
Thank you, C++ is the right choice in this case. I think it's also worth including the new tool into the precompiled packages.

mull_package.bzl
```diff
@@ -11,7 +11,8 @@ load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
 load("@rules_shell//shell:sh_test.bzl", "sh_test")
 load(":mull_publish.bzl", "mull_publish_script")
 
-EXPECTED_MACOS_PACKAGE_CONTENT = """bin/mull-reporter-{LLVM_VERSION}
+EXPECTED_MACOS_PACKAGE_CONTENT = """bin/mull-instrument-{LLVM_VERSION}
```
ccotter, line 14: I'm not actually sure if this is sufficient, or more is needed for the precompiled packaging. I'm not familiar with how it works.

AlexDenisov: That's all it takes really!

ccotter:
@AlexDenisov - could you please cut a new release too?

AlexDenisov:
On it, slipped through the cracks https://github.com/mull-project/mull/pull/1187

Issue #1186, opened 2026-04-23
cxx_replace_scalar_call replaces bool with invalid APInt(1, 42)


ccotter:
Mutating the following code with cxx_replace_scalar_call constructs APInt(1, 42), which hits an assertion in APInt's ctor in debug mode. This doesn't assert in an non-debug build.

```
bool foo(void);
void bar(void) {
    bool x = foo();
}
```

I think we can either use APInt's `implicitTrunc` option, or introduce a `BoolCallReplacement` to be more explicit. Any preferences?

AlexDenisov:
This ties in into the other relevant issue/proposal: https://github.com/mull-project/mull/issues/1142
I'd prefer addressing 1142 as a better fix for this issue, but I think `implicitTrunc` is a good option either way.

ccotter:
Ok, I think #1142 would be a better fix. The assert only triggers for non-release builds anyway, so my specific issue is not all that urgent.

AlexDenisov:
As I'm looking into #1142, I think I mixed up different issues. For call replacements, there is nothing to flip as we are picking a "random" constant value.

AlexDenisov:
And it seems there is no truncation out of the box for older versions of LLVM https://github.com/mull-project/libirm/pull/36 :sigh:

AlexDenisov:
Actually, it was truncated by default in LLVM 19 and below, since LLVM 20 the default is different, so explicit truncation must be requested.

PR #1192, opened 2026-09-10
Negate boolean stores in cxx_assign_const/cxx_init_const

### LINKED ISSUE #1191: cxx_assign_const/cxx_init_const generate equivalent mutants on bool (42 truncates to false)
### Summary

`cxx_assign_const` and `cxx_init_const` replace a stored value with the constant `42` regardless of the store's type. For `bool` this is degenerate: a `bool` is stored as `i8` but read back via `trunc i8 %v to i1`, which keeps only the low bit. `42` is even, so the mutated value always reads back as `false`.

Mutating `bool x = false` to `bool x = 42` therefore produces a program **semantically identical to the original** — an equivalent mutant that no test can kill. It is reported as surviving and depresses the mutation score.

### Reproducer

`main.c`:
```c
#include <assert.h>
#include <stdbool.h>

int main() {
    bool flag = true;
    assert(flag);

    bool negative_flag = false;
    assert(!negative_flag);

    return 0;
}
```

`mull.yml`:
```yaml
mutators:
  - cxx_assign_const
  - cxx_init_const
junkDetectionDisabled: true
```

```
clang -O0 -fpass-plugin=mull-ir-frontend-19 -grecord-command-line -g main.c -o main.exe
mull-runner --allow-surviving --ide-reporter-show-killed -reporters=IDE ./main.exe
```

### Actual

```
[info] Killed mutants (2/4):
main.c:5:10: warning: Killed: Replaced with = 42; [cxx_assign_const]
main.c:5:10: warning: Killed: Replaced with 42 [cxx_init_const]
[info] Survived mutants (2/4):
main.c:8:10: warning: Survived: Replaced with = 42; [cxx_assign_const]
main.c:8:10: warning: Survived: Replaced with 42 [cxx_init_const]
[info] Mutation score: 50%
```

### Expected

The two survivors on line 8 should be killable, or should not be generated. The reported 50% implies a coverage gap, but this test suite is complete with respect to killable mutants: of the 4 reported, only the 2 on line 5 can be killed, and both are.

### Why

Clang emits, at `-O0`:

```llvm
store i8 1, ptr %flag            ; bool flag = true
%4 = load i8, ptr %flag
%5 = trunc i8 %4 to i1           ; only the low bit is read
...
store i8 0, ptr %negative_flag   ; bool negative_flag = false
%10 = load i8, ptr %negative_flag
%11 = trunc i8 %10 to i1
```

Both mutators rewrite the store to `store i8 42`. Since `trunc i8 42 to i1 == 0`:

| Line | Original | Mutated | Assert | Result |
|---|---|---|---|---|
| `bool flag = true` | `true` | `false` | `assert(flag)` fails | Killed |
| `bool negative_flag = false` | `false` | `false` | `assert(!negative_flag)` passes | **Survived (equivalent)** |

Verified independently: storing 42 into a `bool`'s storage and reading it back yields `0`; storing 43 (low bit set) yields `1`.

### Root cause

`lib/Mutators/CXX/NumberMutators.cpp:10`

```cpp
mutators.emplace_back(new irm::StoreIntReplacement(42));
```

The constant is type-agnostic. `getNumberMutators()` is shared by `NumberAssignConst` (`:20`) and `NumberInitConst` (`:29`), which is why both mutators are affected — confirmed in the run above.

### Suggested fix

Make the replacement bool-aware: for a store into an `i1`/`bool`, negate the stored constant (`true`↔`false`) rather than using 42. That yields a genuinely killable mutant in both directions. Alternatively, skip `bool` stores for these mutators, since "replace with 42" has no meaningful reading for a boolean.

Yaraslaut:
Fixes #1191.

`cxx_assign_const` and `cxx_init_const` replace a stored value with the constant 42 regardless of the store's type. A `bool` is stored as a byte but read back through `trunc iN %v to i1`, which keeps only the low bit. 42 is even, so `bool x = false` mutated to `x = 42` still reads as `false` — a program identical to the original, an equivalent mutant no test can kill. It is reported as surviving and lowers the mutation score.

### The change

Stores of the constants 0 or 1 into a location that is read as a boolean are negated instead of set to 42.

The `trunc` is what distinguishes a bool from the other byte-sized types. At IR level:

```llvm
bool b = false;   store i8 0  /  load i8  /  trunc i8 %3 to i1
char c = 0;       store i8 0  /  load i8  /  icmp ne i8 %3, 0
char d = 0;       store i8 0  /  load i8  /  icmp sgt i32 %4, 5   ; d > 5
```

The stores are byte-identical; only the bool is truncated to `i1`. Detecting the `trunc` therefore classifies precisely the stores for which replacing the value with 42 is degenerate, and it needs neither debug info nor LLVM-version-conditional code. Stores this cannot classify keep the 42 replacement, so no other type changes behaviour — including the compiler-synthesized Itanium static-initialization guard variable, which is compared with `icmp eq`, not truncated.

Before, on the reproducer from #1191:

```
Killed (2/4):   main.c:5:10 Replaced with = 42;  [cxx_assign_const]
Survived (2/4): main.c:8:10 Replaced with = 42;  [cxx_assign_const]   <- equivalent, unkillable
Mutation score: 50%
```

After:

```
Killed (6/6):   main.c:5:8  Replaced with = 0;   [cxx_assign_const]
                main.c:8:8  Replaced with = 1;   [cxx_assign_const]
                main.c:13:8 Replaced with = 42;  [cxx_assign_const]   <- char, unchanged
Mutation score: 100%
```

### Report text

The replacement value now varies per mutation, so `MutationPoint` carries its own replacement text, defaulting to the mutator's. Mutators that never produce a boolean mutation are unaffected and keep their fixed string.

Boolean mutants report `= 1;` / `= 0;` rather than `true` / `false`. The patches reporter splices this text directly into generated source (`rust/mull-reporters/src/patches_reporter.rs`), and `1`/`0` is valid in both C and C++ — `true` would not compile in a pre-C23 `.c` file without `<stdbool.h>`, and mull mutates C.

### Tests

- `tests/unit/Mutators/IntegerStoreReplacementTests.cpp` — builds the IR in memory and covers both negation directions, the `char` case (a byte-sized store read via `icmp` keeps 42), a store never read, wider integers, and a non-boolean constant in a boolean slot.
- `tests/integration/mutations/assignment/cxx_assign_const/02/` — end-to-end on the reproducer, covering `bool` alongside every type that resembles one in IR: `char`, `unsigned char`, `signed char`, `short` and `int`, each with a 0 or 1 initializer. All 14 mutants are killed; the bools report `= 0;`/`= 1;` and the rest report `= 42;`.

That coverage is deliberate: an earlier draft of this fix keyed off the constant's bit width instead of the `trunc`, which negated `char` stores too. That turns a killable mutant into an equivalent one — `signed char a = 0; ... if (a > 5)` is killed by the 42 mutant but not by a 1 mutant. Both test files fail against that approach; forcing the classifier to accept every store fails four of the unit tests, including the `char` and static-guard cases.

Full suite passes on LLVM 19 (255 passed, 2 skipped — the arm64-only e2e variants). No existing test needed updating: every current test of these two mutators uses `int` or a non-constant operand.

### Not addressed here

- `bool b = (x > y);` is a non-constant store, so it still gets 42 and remains equivalent. Fixing it means emitting a negation instruction rather than substituting a constant.
- `cxx_replace_scalar_call` has the same truncation problem for bool-returning calls (`irm::IntCallReplacement(42)` on an `i1` folds to `false`) and is untouched.
- The clang AST frontend still hardcodes 42 and gates on `IntTy`/`FloatTy`/`DoubleTy`, so it does not mutate bools at all. That divergence predates this change.
- A template instantiated with both `bool` and `int` can now yield two mutants at one source location whose encodings differ only in replacement text, where previously both encoded `42`. `Mutant`'s `Hash`/`Eq` use only the identifier, so such a pair no longer dedups.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


AlexDenisov:
Hi @Yaraslaut, thank you for putting the effort into fixing this. I believe this is a move into the right direction, but I'm not sure if this is exactly the right approach and the right place for the fix.

This issue has been reported already https://github.com/mull-project/mull/issues/1186 and there is a much better solution proposed, which I didn't have a chance of implementing yet https://github.com/mull-project/mull/issues/1142.

